PR #46, opened 2016-01-06
m3u8 file support！

ai930:
m3u8 file support！


diegomarcon:
m3u8 format was added on PR #52

PR #52, opened 2016-01-13
Support live streaming

diegomarcon:


Pod/Classes/Base/MediaControl.swift
```diff
@@ -143,7 +165,25 @@ public class MediaControl: UIBaseObject {
...
+        updateBars()
+        updateScrubberPosition()
+        updatePlaybackControlButtonIcon()
+    }
+    
+    private func setupForLive() {
+        seekPercentage = 1
+        progressBarView.backgroundColor = UIColor.redColor()
+        durationLabel.text = ""
+        currentTimeLabel.text = ""
```
barbosa: Hey @diegomarcon,

there is a tiny white separator between the `durationLabel` and the `currentTimeLabel`. Do you hide it when dealing with live media?

Pod/Classes/Base/MediaControl.swift
```diff
@@ -217,7 +263,7 @@ public class MediaControl: UIBaseObject {
         hideControlsTimer.invalidate()
     }
     
-    @IBAction func handleScrubberPan(panGesture: UIPanGestureRecognizer) {
+    @IBAction func handleScrubberPan(panGesture: UIPanGestureRecognizer) {
```
barbosa: There is an extra `tab` here.

Pod/Classes/Base/MediaControl.swift
diegomarcon: @barbosa 

Actually I was not hiding it. It's a discreet separator but it does look odd when paying attention

r-a-o:
Awesome! Just want to know if there is a way to provide multiple stream urls so player can pick based on network bandwidth?
Is this available in current implementation for live and on-demand videos? If its not currently present, I am willing to implement it, any pointer on this would help immensely. 

Thanks!

PR #57, opened 2016-03-11
BaseObject: Add eventKey for handling on/off events

diegomarcon:
Why? 
- Avoid using callback memory address for identifying event.


towerz:
Good job @diegomarcon! Thanks!

PR #61, opened 2016-04-12
Allow playback replacement

diegomarcon:


diegomarcon:
Whoops. We should not allow playbacks to be replaced. 
Container should have a load source and select a playback that knows how to play it.
:hankey:

PR #63, opened 2016-04-28
Player: Events listening interface

diegomarcon:


towerz:
@diegomarcon I'd rather not make external event callbacks fall through directly to the container.

It may seem weird, but it would be better if the Player class extends BaseObject, exposing PlayerEvents as actual events. This way we avoid coupling the external interface to the container/playback internal interface, as well as exposing inner behavior.

PR #67, opened 2016-05-23
MediaControl extensibility improvements

towerz:
Changed MediaControl elements to explicit optionals, so we can safely ignore missing elements when loading custom XIBs.


diegomarcon:
👍

PR #74, opened 2016-06-16
Add subtitle and audio source abstractions

### LINKED ISSUE #72: Create Subtitle and AudioSource abstractions
Currently in clappr we are using AVFoundation abstractions for subtitles and audiosources.

diegomarcon:
Clappr should not use models from AVFoundation.

Closes #72

### DIFF
--- a/Pod/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Pod/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -18,19 +18,18 @@ public class AVFoundationPlayback: Playback {
     public override var pluginName: String {
         return "AVPlayback"
     }
-
+    
     public override var selectedSubtitle: Subtitle? {
         get {
             let option = getSelectedMediaOptionWithCharacteristic(AVMediaCharacteristicLegible)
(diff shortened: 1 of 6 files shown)


Pod/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -19,30 +19,38 @@ public class AVFoundationPlayback: Playback {
...
-    public override var selectedSubtitle: AVMediaSelectionOption? {
+    public override var selectedSubtitle: Subtitle? {
         get {
-            return getSelectedMediaOptionWithCharacteristic(AVMediaCharacteristicLegible)
+            let option = getSelectedMediaOptionWithCharacteristic(AVMediaCharacteristicLegible)
+            return Subtitle.fromAVMediaSelectionOption(option)
         }
         set {
-            setMediaSelectionOption(newValue, characteristic: AVMediaCharacteristicLegible)
+            if let newOption = newValue?.raw as? AVMediaSelectionOption {
```
towerz: There should be a way to disable subtitles (setting the `AVMediaSelectionOption` to `nil`)

PR #76, opened 2016-06-20
AVPlayback: Refactor can play to deal with path extension

diegomarcon:
Before this PR AVFoundationPlayback was accepting any valid NSURL as a source.


Pod/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -46,11 +46,13 @@ public class AVFoundationPlayback: Playback {
...
     public override class func canPlay(options: Options) -> Bool {
-        guard let urlString = options[kSourceUrl] as? String, let _ = NSURL(string: urlString) else {
+        guard let urlString = options[kSourceUrl] as? String,
+            let url = NSURL(string: urlString),
+            let pathExtension = url.pathExtension else {
             return false
         }
         
-        return true
+        return AVURLAsset.isPlayableExtendedMIMEType("video/\(pathExtension)")
```
towerz: I'm not sure this would work. Some streaming protocols don't use the `video/` prefix.  For example, m3u8 manifests use `application/x-mpegURL` as MIME type.

towerz:
👍

PR #84, opened 2016-06-23
Add player state properties to public API

### LINKED ISSUE #77: Improve public API to provide more status information
Something like this:
```swift
player.isFullscreen // Bool - whether or not the player is in fullscreen mode
player.version // String - the version string of the player instance

player.activeContainer // Container? - the container that is currently active 
player.activePlayback // Playback? - the playback associated to the active container
player.activePlayback.isPlaying // Bool - whether the active playback is playing the video
player.activePlayback.isPaused // Bool - whether the active playback is in a paused state
player.activePlayback.isBuffering // Bool - whether the active playback is in a buffering state
player.activePlayback.duration // Double - the duration of the video, in seconds
player.activePlayback.position // Double - the current position of the playback, in seconds
```

towerz:
Also refactors most core/container/playback states to properties. (Closes #77)

### DIFF
--- a/Pod/Classes/Base/Core.swift
+++ b/Pod/Classes/Base/Core.swift
@@ -19,6 +19,10 @@ public class Core: UIBaseObject, UIGestureRecognizerDelegate {
     public var activePlayback: Playback? {
         return activeContainer?.playback
     }
+
+    public var isFullscreen: Bool {
+        return self.fullscreenController.presentingViewController != nil
+    }
     
     required public init?(coder aDecoder: NSCoder) {
         fatalError("Should be using init(sources:[NSURL]) instead")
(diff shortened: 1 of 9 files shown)


Pod/Classes/Base/Core.swift
```diff
@@ -8,6 +8,17 @@ public class Core: UIBaseObject, UIGestureRecognizerDelegate {
     public var parentView: UIView?
     private var loader: Loader
     private lazy var fullscreenController = FullscreenController(nibName: nil, bundle: nil)
+
+    public var activeContainer: Container? {
+        guard let container = self.container else {
+            return nil
+        }
+        return container
```
diegomarcon: Can't we remove the guard clause? activeContainer is already optional

towerz: Indeed 👍

PR #97, opened 2016-07-21
Refactor plugin loading

fcapinho:


Pod/Classes/Plugin/Container/LoadingContainerPlugin.swift
```diff
@@ -1,12 +1,9 @@
 public class LoadingContainerPlugin: UIContainerPlugin {
     
-    private var spinningWheel: UIActivityIndicatorView
+    private var spinningWheel: UIActivityIndicatorView?
```
towerz: I think it would be better to use an implicitly unwrapped optional here:
```swift
private var spinningWheel: UIActivityIndicatorView!
```

Pod/Classes/Plugin/Container/LoadingContainerPlugin.swift
```diff
@@ -26,32 +30,36 @@ public class LoadingContainerPlugin: UIContainerPlugin {
         translatesAutoresizingMaskIntoConstraints = false
         
         let widthConstraint = NSLayoutConstraint(item: self, attribute: .Width, relatedBy: .Equal,
-            toItem: nil, attribute: .NotAnAttribute, multiplier: 1, constant: spinningWheel.frame.width)
+                                                 toItem: nil, attribute: .NotAnAttribute, multiplier: 1,
```
towerz: Just for style purposes, when this happen we should format with one of these options:
- If there are no blocks in the parameter list, we should avoid column indentation, only indenting by one level, like it was before this change.
```swift
let widthConstraint = NSLayoutConstraint(item: self, attribute: .Width, relatedBy: .Equal,
    toItem: nil, attribute: .NotAnAttribute, multiplier: 1,
    constant: spinningWheel!.frame.width)
```
- If there are blocks in the parameter list, or we want a column-style formatting for some reason, break the line on the first parameter (just after `(`) and then put a parameter per line, with a single indentation for each
```swift
let widthConstraint = NSLayoutConstraint(
    item: self, 
    attribute: .Width,
    relatedBy: .Equal,
    toItem: nil, 
    attribute: .NotAnAttribute,
    multiplier: 1,
    constant: spinningWheel!.frame.width)
```

Pod/Classes/Plugin/Container/UIContainerPlugin.swift
```diff
@@ -16,6 +16,15 @@ public class UIContainerPlugin: UIPlugin, Plugin {
         super.init(frame: CGRectZero)
     }
     
+    public required init(context: UIBaseObject) {
+        super.init(frame: CGRectZero)
+        if let tempContainer = context as? Container {
```
towerz: The `container` attribute should be an implicitly unwrapped optional, to avoid force unwrapping on calls and chaining `?`. We already check for `nil` instances on the initializer.
When using `if let` or `guard let` statements, we should keep the variable name clean of `temp` prefixes. Using just `container` as variable name would be better:
```swift
if let container = context as? Container {
    self.container = container
}
```
We should avoid using `;` as well, unless required.

Pod/Classes/Plugin/Container/UIContainerPlugin.swift
```diff
@@ -16,6 +16,15 @@ public class UIContainerPlugin: UIPlugin, Plugin {
...
+            self.container = tempContainer;
```
towerz: We should avoid using `;` as well, unless required.

Pod/Classes/Plugin/Core/UICorePlugin.swift
```diff
@@ -16,6 +16,15 @@ public class UICorePlugin: UIPlugin, Plugin {
         super.init(frame: CGRectZero)
     }
 
+    public required init(context: UIBaseObject) {
+        super.init(frame: CGRectZero)
+        if let tempCore = context as? Core {
```
towerz: Same thing here: use implicitly unwrapped optional and avoid using `temp` prefixes.

towerz:
🚀

PR #99, opened 2016-07-25
Support Carthage

diegomarcon:


towerz:
@diegomarcon I don't think we should include Carthage dependencies on our repository. Instead, we should either ignore those (since Carthage checks them out by itself) or use submodules instead.

diegomarcon:
@towerz Agree. Also removed 'Pods' folder that was added by mistake

towerz:
:beers:

PR #103, opened 2016-09-05
MediaOption improvements

towerz:
MediaOption is now Equatable, and the default subtitle option is `offSubtitle` instead of `nil`.


diegomarcon:
🍻

PR #111, opened 2017-03-15
Migration to Swift 3

brunopinheiro:
This branch will fail the CI pipeline until it is not done.
Swift 3 Migration should be completed before we can get a valid CI result


gabiduarte:
[Migrar Clappr para Swift 3](https://trello.com/c/EInVKkOM/24-migrar-clappr-para-swift-3)

PR #113, opened 2017-03-16
Refactor container plugin creation

diegomarcon:
This PR removes the container factory which created the playback before initializing container plugins.
Now, the container is responsible for loading the source and adding the plugins.


Tests/PosterPluginTests.swift
```diff
@@ -9,9 +9,7 @@ class PosterPluginTests: QuickSpec {
             var container: Container!
             let options = [kSourceUrl : "http://globo.com/video.mp4",
                            kPosterUrl : "http://clappr.io/poster.png"]
-            let playback = StubPlayback(options: options)
```
diegomarcon: 👍

PR #115, opened 2017-03-28
Enable Background Playback

brunopinheiro:
Changes to enable the external playback keep running on background


gabiduarte:
[Habilitar Playback Externo em Background](https://trello.com/c/LVRNIZ5n/34-habilitar-playback-externo-em-background)

Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -5,350 +5,359 @@ enum PlaybackState {
 }
 
 open class AVFoundationPlayback: Playback {
```
diegomarcon: Shouldn't we be using 4 spaces for tabs?

brunopinheiro: Just pushed a new commit fixing the tabs 😄

diegomarcon: 👍

PR #121, opened 2017-05-09
Implement destroy logic

igoriols:
* Add destroy method for playback, plugins, container, core, media control and player.


gabiduarte:
[Player tocando em Background #16](https://trello.com/c/lCc1q34b/64-player-tocando-em-background-16)

Issue #12, opened 2014-11-19
Create the playback abstraction


towerz:
The playback abstraction is responsible for rendering and playing the video.

PR #125, opened 2017-06-16
fix: pause on background issue

geisonfgf:


brunopinheiro:
[NPAW #12: Eventos de /start e /stop ao entrar em background](https://trello.com/c/M3OU6cnO/123-npaw-12-eventos-de-start-e-stop-ao-entrar-em-background)

PR #127, opened 2017-07-19
Remove previous Playback from view

brunopinheiro:
When the user is trying to load a new resource, the previous playback was not removed from the container's view, causing the player to keep showing the last video's frame, while the new video is being played on background.

### DIFF
--- a/Clappr/Classes/Base/Container.swift
+++ b/Clappr/Classes/Base/Container.swift
@@ -52,7 +52,7 @@ open class Container: UIBaseObject {
         playbackOptions[kSourceUrl] = source
         playbackOptions[kMimeType] = mimeType
 
-        self.playback?.removeFromSuperview()
+        self.playback?.destroy()
 
         let playbackFactory = PlaybackFactory(loader: loader, options: playbackOptions)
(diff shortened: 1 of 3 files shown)


brunopinheiro:
[Congelamento de imagem ao utilizar load() #24](https://trello.com/c/EwV24O0Y/170-congelamento-de-imagem-ao-utilizar-load-24)

Clappr/Classes/Base/Container.swift
```diff
@@ -52,8 +52,9 @@ open class Container: UIBaseObject {
         playbackOptions[kSourceUrl] = source
         playbackOptions[kMimeType] = mimeType
 
-        let playbackFactory = PlaybackFactory(loader: loader, options: playbackOptions)
+        self.playback?.removeFromSuperview()
```
igoriols: @brunopinheiro I think it would be better to call `self.playback?.destroy()` to destroy the old playback instance. Can you, please, check if it works?

brunopinheiro: You're right! It makes more sense to use destroy. We're gonna push a new commit

PR #133, opened 2017-08-21
Fix player slider swipe when inside a ScrollView

nabilsafatli:
**Goal**
A problem was happening when the player was inside a scrollview, the conflict was happening between the swipe and scroll events.
We intend to change the MediaControl slide bar for a native component.

**How to test**

To simulate the problem, open the sample and put the player inside a ScrollView.


gabiduarte:
[Conflito no slide do vídeo](https://trello.com/c/G5ZCz771/225-conflito-no-slide-do-v%C3%ADdeo)

PR #137, opened 2017-08-30
Prevent pause event after video is ended

igoriols:
## Goal

The problem is that when video is ended, a pause event is being triggered. To fix that, we prevent change the internal state of the playback after the video is ended.

## How to test

Run the clappr sample and observe if an "on Paused" is printed on the log after the video is ended.


CristianDRM:
[Pause após o video finalizar](https://trello.com/c/1qbvCDWu/344-pause-ap%C3%B3s-o-video-finalizar)

nabilsafatli:
@igoriols , @fcapinho , @CristianDRM  should we close this?

igoriols:
The problem doesn't occurs anymore.

PR #138, opened 2017-09-26
Fullscreen handle by application

CristianDRM:
### Summary
When user requests fullscreen, the player change the layout to fullscreen by itself.

### Goal
 - Create an option that will be used when the application needs to handle with fullscreen instead of player.
 - Fire an event to notify app that user request's fullscreen or exitFullscreen
 - Add option to test the option on ``Sample``

#### Test: Application controls fullscreen
 1. Run App and switch the option **fullscreen controled by app** to **on**.
 2. Open the video
 3. Change the orientation to landscape
 4. **The player should change the icon to fullscreen**
 5. Change the orientation to portrait
 6. **The player should change the icon to default(embed)**

#### Test: Player controls fullscreen
 1. Run App and switch the option **fullscreen controled by app** to **off**.
 2. Open the video
 3. Tap to set the video to fullscreen
 4. **The player should change the layout and icon to fullscreen**
 5. Tap to close fullscreen of video
 6. **The player should change the layout and icon to embed**


CristianDRM:
[Permitir que a app iOS possa controlar o fullscreen](https://trello.com/c/8RTrKeAK/294-permitir-que-a-app-ios-possa-controlar-o-fullscreen)

Clappr/Classes/Base/Options.swift
```diff
@@ -5,9 +5,22 @@ public let kSourceUrl = "sourceUrl"
...
 public let kFullscreenDisabled = "fullscreenDisabled"
+public let kFullscreenByApp = "fullscreenByApp"
 public let kStartAt = "startAt"
 public let kAutoPlay = "autoplay"
 public let kPlaybackNotSupportedMessage = "playbackNotSupportedMessage"
 public let kMimeType = "mimeType"
 public let kDefaultSubtitle = "defaultSubtitle"
 public let kDefaultAudioSource = "defaultAudioSource"
+
+struct OptionsWrapper {
```
geisonfgf: Since seems to me that the functionality of this class is unbox the optional values could make more sense call it OptionsUnboxer?

Clappr/Classes/Base/Player.swift
```diff
@@ -127,10 +127,10 @@ open class Player: BaseObject {
         core?.activePlayback?.seek(timeInterval)
     }
 
-    open func setFullscreen(_ fullscreen: Bool) {
```
fcapinho: I recommend keeping the `setFullscreen` external interface as it is used by all other apps.

PR #143, opened 2017-10-25
Fix setFullscreen when the fullsscreen is controled by player

CristianDRM:
## Goal
After the fullscreen controled by app was added, the method ``setFullscreen`` doesnt set player to fullscreen when **kFullscreenByApp is false**.

## How to test
1) Open Sample
2) Disable **Fullscreen controled by app** and **Start player as Fullscreen**
3) Tap on **start video**
4) Change orientation of App to landscapeLeft

**obs: Make sure that the player was set to fullscreen when orientation changes**

### DIFF
--- a/Clappr/Classes/Base/Core.swift
+++ b/Clappr/Classes/Base/Core.swift
@@ -145,7 +145,8 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     }
 
     open func setFullscreen(_ fullscreen: Bool) {
-        mediaControl?.fullscreen = fullscreen
+        guard let mediaControl = mediaControl, mediaControl.fullscreen != fullscreen else { return }
+        mediaControl.fullscreen = fullscreen
     }
 
     open func destroy() {
--- a/Clappr/Classes/Base/Core.swift
+++ b/Clappr/Classes/Base/Core.swift
@@ -145,12 +145,7 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     }
 
     open func setFullscreen(_ fullscreen: Bool) {
-        guard let mediaControl = mediaControl, mediaControl.fullscreen != fullscreen else { return }
-        if optionsUnboxer.fullscreenControledByApp {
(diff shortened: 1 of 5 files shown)


CristianDRM:
[setFullscreen não funciona](https://trello.com/c/Gw708iEi/342-setfullscreen-n%C3%A3o-funciona)

Clappr/Classes/Base/Core.swift
```diff
@@ -145,7 +145,12 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     }
 
     open func setFullscreen(_ fullscreen: Bool) {
-        mediaControl?.fullscreen = fullscreen
+        guard let mediaControl = mediaControl, mediaControl.fullscreen != fullscreen else { return }
```
leandroalonso: Can we move this part `mediaControl.fullscreen != fullscreen` to a method and name it from what this concept represents?

CristianDRM: Reviewed 👍

Example/ViewController.swift
```diff
@@ -20,6 +20,33 @@ class ViewController: UIViewController {
         player.attachTo(playerContainer, controller: self)
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        NotificationCenter.default.addObserver(self, selector: #selector(rotated), name: .UIDeviceOrientationDidChange, object: nil)
+    }
+
+    func rotated() {
+        if (player?.isFullscreen)! && (UIDevice.current.orientation == .faceUp || UIDevice.current.orientation == .faceDown) {
```
geisonfgf: Could we change it to a guard let in order to remove the bang and so avoiding the possibility of some crash at runtime?

leandroalonso: This code is in our sample app, not in the Clappr itself. Should we really change it?

CristianDRM: @leandroalonso  I agree with you, but i change the implementation to be more clear about what the method does.

PR #144, opened 2017-11-06
Exposes asset property in AVFoundationPlayback

leandroalonso:
# Goal

- To expose `asset` property in `AVFoundationPlayback`

# How to test

- Just make sure that this change didn't affect Clappr functionality


nabilsafatli:
Hey @leandroalonso , why do you need to expose this property?

PR #146, opened 2017-11-06
Feature/increase test coverage

CristianDRM:
## Goal
Increase test coverage for Core.swift and Container.swift.
Fix events triggered when the activeContainer was changed.

### DIFF
--- a/Tests/ContainerTests.swift
+++ b/Tests/ContainerTests.swift
@@ -51,6 +51,10 @@ class ContainerTests: QuickSpec {
                     expect(container.subviews.first) == playback
                 }
 
+                it("should set frame of container as CGRect.zero") {
+                    expect(container.view?.frame) == CGRect.zero
+                }
+
                 it("Should have a constructor that receive options") {
                     let options = ["aOption": "option"]
                     let container = Container(options: options)
--- a/Tests/ContainerTests.swift
+++ b/Tests/ContainerTests.swift
@@ -52,7 +52,7 @@ class ContainerTests: QuickSpec {
                 }
 
                 it("should set frame of container as CGRect.zero") {
-                    expect(container.view?.frame) == CGRect.zero
+                    expect(container.frame) == CGRect.zero
                 }
 
                 it("Should have a constructor that receive options") {
--- a/Tests/ContainerTests.swift
+++ b/Tests/ContainerTests.swift
@@ -5,97 +5,137 @@ import Nimble
 class ContainerTests: QuickSpec {
 
     override func spec() {
-        describe("Container") {
-            let options: Options = [kSourceUrl: "http://clappr.com/video.mp4"]
-            let optionsWithInvalidSource = [kSourceUrl: "invalid"]
+
+        struct Resource {
+            static let invalid = [kSourceUrl: "invalid"]
+            static let valid = [kSourceUrl: "http://clappr.com/video.mp4"]
+        }
+
+        fdescribe(".Container") {
 
             var container: Container!
-            var playback: StubPlayback!
             var loader: Loader!
 
             beforeEach {
                 loader = Loader()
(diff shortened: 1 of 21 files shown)


CristianDRM:
[[Nova UI] Aplicar testes na estrutura atual](https://trello.com/c/HTn1p4Ny/347-nova-ui-aplicar-testes-na-estrutura-atual)

Tests/ContainerTests.swift
```diff
@@ -27,6 +27,14 @@ class ContainerTests: QuickSpec {
                     expect(container.playback?.pluginName) == "AVPlayback"
                 }
 
+                it("should has background color `clear`") {
```
leandroalonso: We don't need to necessarily change this right now, but keep in mind: using `should` in tests is a bad practice.

Quoting [from here](http://www.betterspecs.org/#should):

> Do not use should when describing your tests. Use the third person in the present tense.

CristianDRM: Ok, i will read this on future to improve my unit tests.

Tests/CoreTests.swift
```diff
@@ -126,12 +142,148 @@ class CoreTests: QuickSpec {
...
+
+                    core.on(InternalEvent.willDestroy.rawValue) { _ in
+                        didCallDidDestroy = true
+                    }
+                    core.destroy()
+
+                    expect(didCallDidDestroy).toEventually(beTrue())
+                }
+
+                it("Should remove listeners") {
```
leandroalonso: This test is doing more than what the description informs: it checks for the correctness of the `listenTo` and then check if the listener was removed. We could break this for better legibility.

Tests/CoreTests.swift
```diff
@@ -126,12 +142,148 @@ class CoreTests: QuickSpec {
...
+
+                    core.on(InternalEvent.didChangeActiveContainer.rawValue)   { userInfo in
+                        didCallChangeActiveContainer = true
+                    }
+                    core.activeContainer = Container()
+
+                    expect(didCallChangeActiveContainer).toEventually(beTrue())
+                }
+
+                it("Should listen willChangePlayback and trigger willChangeActivePlayback for new Container") {
```
leandroalonso: This test might give a false positive. Imagine the scenario where no `willChangePlayback` is triggered, but instead, two `willChangeActivePlayback` was triggered. The test will pass, but the behaviour is incorrect.

We should check if every callback was called, not the number of times.

Tests/CoreTests.swift
```diff
@@ -126,12 +142,148 @@ class CoreTests: QuickSpec {
...
+
+                    core.activeContainer = container
+                    core.activeContainer?.playback = AVFoundationPlayback()
+                    core.activeContainer = Container() // prevent events to be called for old container
+
+                    expect(countOfCallEvents).toEventually(equal(2))
+                }
+
+                it("Should listen didChangePlayback and trigger didChangeActivePlayback for new Container") {
+                    let core = Core()
```
leandroalonso: Same as before, it could give a false positive.

Tests/CoreTests.swift
```diff
@@ -126,12 +142,148 @@ class CoreTests: QuickSpec {
...
+
+                    core.activeContainer = container
+                    core.activeContainer?.playback = AVFoundationPlayback()
+                    core.activeContainer = Container() // prevent events to be called for old container
+
+                    expect(countOfCallEvents).toEventually(equal(2))
+                }
+
+                it("Should'nt trigger events twice when container and playback was changed") {
+                    let core = Core()
```
leandroalonso: Again. :)

Tests/CoreTests.swift
CristianDRM: I remove the trigger test, cuz doesnt make sense to test this action here.

PR #152, opened 2017-12-11
Automate tests and release

leandroalonso:
# Goal

To automate tests and release of Clappr.

# How to test

## Unit and UI tests

1. `make setup` and `make test`: makes sure that all the tests pass

## Release

**This is just an explanation of how the release will work. You should not try this unless you're really publishing a new Clappr version.**

1. You need write permissions on `clappr-ios` repo, otherwise, the release will not work;
2. Makes sure to put your Github token in your profile. You can find instructions on [how to generate a token here](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/). Then, you have to put the token in your profile (`.bash_profile` or `.zshrc`), like this:

```
export GITHUB_TOKEN="your-token-goes-here"
```
3. Make sure that your repository is updated (mainly `dev` and `master` branches)
4. `make release version=x.x.x`
5. Checks if a new release draft was created on Github and published on Cocoapods

# To discuss

1. Should we rely on a `CHANGELOG` to publish the new Clappr version? If yes, how to handle that? You change the `CHANGELOG` and then commit the result and do a release?
2. Should we update all necessary branches automatically? (`master` and `dev`)


leandroalonso:
Apparently `Travis CI` uses `RVM`. This causes an incompatibility with `Rbenv` and our build now is broken. We have a few alternatives:

- Revert `.travis.yml`. This will fix the build but Travis will run our build in a different way on how we're going to do on our machines;
- Search for a new CI solution to use

What do you think? Any other idea?

PR #154, opened 2017-12-20
Remove autoplay option

CristianDRM:
### Goal
Remove kAutoplay option


CristianDRM:
[Remover a option de KAutoPlay do Clappr](https://trello.com/c/mSQYZmZp/411-remover-a-option-de-kautoplay-do-clappr)

PR #157, opened 2017-12-27
UIPlugin now conforms with UIObject

leandroalonso:
# Goal

As we're evolving our architecture our first step is to make `UIPlugin` conforming to `UIObject`, having a `View?` and a `render()` method

# How to test

Makes sure all tests pass.

### DIFF
--- a/Clappr/Classes/Base/UIObject.swift
+++ b/Clappr/Classes/Base/UIObject.swift
@@ -1,5 +1,5 @@
 protocol UIObject {
-    var view: UIView? { get set }
+    var view: UIView? { get }
 
     func render()
 }
--- a/Clappr/Classes/Base/UIObject.swift
+++ b/Clappr/Classes/Base/UIObject.swift
@@ -1,0 +1,5 @@
+protocol UIObject {
+    var view: UIView? { get set }
+
+    func render()
+}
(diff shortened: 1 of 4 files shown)


leandroalonso:
[[Nova UI] Criar MediaControl como plugin no Steve](https://trello.com/c/1kgTo6MV/5-nova-ui-criar-mediacontrol-como-plugin-no-steve)

Clappr/Classes/Base/UIObject.swift
```diff
@@ -0,0 +1,5 @@
+protocol UIObject {
+    var view: UIView? { get set }
```
CristianDRM: I think that is better to just provide get in order to set the concrete implementation as private(set), what you think?

Tests/UIPluginTests.swift
```diff
@@ -5,45 +5,21 @@ import Nimble
 class UIPluginTests: QuickSpec {
 
     override func spec() {
-        describe("Instantiation") {
+        describe("UIPlugin") {
```
CristianDRM: Following the project conventions, the Class name on describe must have the prefix: ``.``

Clappr/Classes/Base/UIObject.swift
leandroalonso: I agree! :)

I will change it, tks. 👍

Tests/UIPluginTests.swift
leandroalonso: We use `.` for class methods (aka `static`) and `#` for instance methods. The class name is neither of these ones... what do you think?

CristianDRM: ;0I ended up confusing the conventions so you can disregard my comment :)

PR #158, opened 2017-12-27
Ping UPA of Apple with now playing

CristianDRM:
### Goal
Add ``Content identifier`` and ``Playback Progress``  for NowPlaying options.
Add options for the properties above and pass it to the external data on AVPlayerCurrentItem.

### How to test

1. Run all unit tests.
2. Open ViewController.
3. Pass the following options to player:

```swift 
let metadata = [kMetaDataTitle: "Foo",
                        kMetaDataDescription: "Lorem ipsum lorem",
                        kMetaDataContentIdentifier: "Foo v2",
                        kMetaDataDate: Date(),
                        kMetaDataWatchedTime: 1] as [String : Any]
let options = [
            kSourceUrl: "https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_ts/master.m3u8",
            kPosterUrl: "https://cloud.githubusercontent.com/assets/1156242/16349649/54f233e2-3a30-11e6-98e4-42eb5284b730.png",
            kAutoPlay: true,
            kMediaControl: true,
            kMetaData: metadata,
            ] as [String: Any]
```
4. Run the project.
5. On remote control, Swipe down and check if the properties of ``more Info`` is the same that you pass on options.
nowplaying


Tests/AVFoundationNowPlayingBuilderTests.swift
```diff
@@ -0,0 +1,257 @@
...
+                context("When doesn't have metadata image") {
+
+                    beforeEach {
+                        metadata = [:]
+                        nowPlayingBuilder = AVFoundationNowPlayingBuilder(metadata: metadata!)
+                    }
+
+                    it("loads poster image") {
+                        waitUntil() { done in
+                            let url = "https://cloud.githubusercontent.com/assets/1156242/16349649/54f233e2-3a30-11e6-98e4-42eb5284b730.png"
```
fcapinho: Network access shall be avoided in unit tests.

CristianDRM: Its necessary to add a framework to stub this request. Seems better to not add new dependencies on Clappr to avoid compatibility problems.

Do you have another suggestion for this problem?

fcapinho: I consider that we can add this new framework as internal testing only.

Clappr.xcodeproj/project.pbxproj
```diff
@@ -402,7 +409,7 @@
 		96D362B21D41339400CCB866 /* Playback */ = {
 			isa = PBXGroup;
 			children = (
-				96D362B31D41339400CCB866 /* AVFoundationPlayback.swift */,
+				9E18A7401FF5512E00C706F4 /* AVFoundation */,
```
jchahoud: @CristianDRM not sure if AVFoundation should be used in a Clappr's folder name as it might cause confusion with Apple's framework name. What do you think?

Clappr/Classes/Plugin/Playback/AVFoundation/AVFoundationNowPlayingBuilder.swift
```diff
@@ -0,0 +1,105 @@
+import MediaPlayer
+import AVKit
+import AVFoundation
+
+class AVFoundationNowPlayingBuilder: AVFoundationNowPlayingBuilderProtocol {
```
jchahoud: @CristianDRM any particular reason for `AVFoundationNowPlayingBuilder` being a class implementing a protocol. Could it be just a struct?

Tests/AVFoundationNowPlayingBuilderTests.swift
leandroalonso: I agree with Fernando. I don't see any problem in adding internal frameworks.

Clappr/Classes/Plugin/Playback/AVFoundation/AVFoundationNowPlayingBuilder.swift
leandroalonso: We've called this `AVFoundationNowPlayingBuilder` and my first thought was that class would build a `AVFoundationNowPlaying` object or something like that. But this is not what's happening.

Can you help me understand what this class is building? Or should we rename it?

CristianDRM: Before [this commit](https://github.com/clappr/clappr-ios/pull/158/commits/d6ba0c613e0635b35a1d56d508d91767b1e3f6fd), the responsibility of this class was just built a list of items.
Now the responsibility of this class is:
* Build a list of items
* Set these items to the AVPlayerItem

Do you know a good name for this struct?

PR #163, opened 2018-01-10
Update PodSpec to support tvOS and iOS

CristianDRM:
### Goal
Adjust Clappr to be published in Cocoapods(iOS and tvOS) with the same Podspec.
### How to test
Just run `pod lib lint` and check if everything is working

### DIFF
--- a/Clappr.podspec
+++ b/Clappr.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name             = "Clappr"
   s.version          = "0.7.3"
-  s.summary          = "An extensible media player for iOS"
+  s.summary          = "An extensible media player for iOS and tvOS"
   s.homepage         = "http://clappr.io"
   s.license          = 'MIT'
   s.authors          = {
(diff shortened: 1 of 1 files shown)


CristianDRM:
[[iOS&tvOS | Clappr] Ajustar podspec](https://trello.com/c/iaWX4HBc/32-iostvos-clappr-ajustar-podspec)

Clappr.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name             = "Clappr"
   s.version          = "0.7.3"
-  s.summary          = "An extensible media player for iOS"
+  s.summary          = "An extensible media player for iOS and tvOS"
```
jchahoud: it seems that the previous Clappr for tvOS version was 0.8.1 as you can check in the related branch podspec https://github.com/clappr/clappr-ios/blob/tvos/Clappr.podspec. Maybe we can use 0.9 for the multi-platform podspec version

nabilsafatli: Hey guys, shouldn't we update only when we are in a release process?
But I totally agree that it can become `0.9` when we release the new version 😄

PR #164, opened 2018-01-11
Bring iOS_tvOS integration branch to Dev

jchahoud:
## Goal

Bring iOS & tvOS Integration Branch to Dev

## How to test

- Run Clappr iOS Example App
- Run Clappr tvOS Example App
- Run all tests


igoriols:
[[iOS&tvOS | Clappr] Integrar feature em Dev](https://trello.com/c/UK2EDimm/37-iostvos-clappr-integrar-feature-em-dev)

igoriols:
Hey guys, there is a warning message related with the name of tvOS target: invalid character '_' in Bundle Identifier at column 17. This string must be a uniform type identifier (UTI) that contains only alphanumeric (A-Z,a-z,0-9), hyphen (-), and period (.) characters.

Do you think we need to remove the "_" and use something else?

Clappr.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name             = "Clappr"
-  s.version          = "0.7.5"
-  s.summary          = "An extensible media player for iOS"
+  s.version          = "0.9.0"
```
leandroalonso: What is the reason to change the framework version?

nabilsafatli: Hey guys,
I asked the same thing on a [previous PR](https://github.com/clappr/clappr-ios/pull/163/files/320efe7f0c7a3da3758f4cbb6791d94de5d7f19f), but it got accepted and merged without an answer. 😅

igoriols: I agree with you guys. In tvOS we didn't publish Clappr in cocoapods, we just used a fixed version in podspec. I think we can change the version to `0.7.5` to fix this change and upgrade it to `0.8.0` when we launch the next version.

What do you think?

leandroalonso: Yep! I would like to keep the current version (`0.7.5`), then we can discuss the version of the next release.

igoriols: Done! :D

PR #167, opened 2018-01-17
Change the order of how Container and Core views are rendered

leandroalonso:
# Goal

We have identified that the Container and Core views were not in the right order, as specified in [our documentation](https://github.com/clappr/clappr-docs/wiki/Architecture).

As you can see, the first layer should be Core and the Container is right behind it. In the current implementation, is the opposite.

# How to test

1. Make sure that all tests are passing;
2. Run Clappr and check if everything behaves correctly


igoriols:
[[Nova UI] Criar botão de play do MediaControl](https://trello.com/c/ypHWgsAw/30-nova-ui-criar-bot%C3%A3o-de-play-do-mediacontrol)

PR #169, opened 2018-01-22
Reorganize files to be more maintainable

CristianDRM:
### Current
Structure of folders and files:
screen shot 2018-01-22 at 11 40 02

Podspec:
screen shot 2018-01-22 at 11 40 36

### After the changes:

Structure of folders and files:
screen shot 2018-01-22 at 11 43 11

Podspec:
screen shot 2018-01-22 at 11 43 35

#### Advantages:
In the current structure, if a new class is added only in the iOS target, it's mandatory to add on exclude_files. 
In the new structure, It's not necessary to add the reference on the podspec.
The files that only exist in iOS will be more visible

### How to Test
Run unit tests and check if everything is ok.


CristianDRM:
[[iOS&tvOS | Clappr] Reorganizar pastas e arquivos do projeto](https://trello.com/c/f8EHa9Tz/42-iostvos-clappr-reorganizar-pastas-e-arquivos-do-projeto)

PR #170, opened 2018-01-30
Decouple Fullscreen logic of Core and MediaControl

CristianDRM:
## Summary
**OBS: Changes only for iOS.**

At the moment the complexity of fullscreen logic is spread between Core and MediaControl.

When the option ``kFullscreenByApp: true`` means that the Clappr doesn't set the player to fullscreen when the user taps on the fullscreen button, instead it triggers an event to notify the desire of the user. Then the app takes the responsibility to put the player in the fullscreen mode.

The Core and MediaControl class hold the logic to know when its necessary to only trigger event, change the icon and change the layout state to embed or fullscreen.

## Goal
Simplify and decouple the logic of fullscreen.
### MediaControl
Media controls don't hold any logic of fullscreen controlled by the app.

#### When user taps on fullscreen button the MediaControl triggers the event: 
 - userRequestExitFullscreen
 - userRequestEnterInFullscreen
#### When core triggers didExitFullscreen/didEnterFullscreen
 - Changes the icon to the respective state

### Core
Now Core holds both logic of fullscreen: Controlled by App or by the Player.

## How to test
1. Run all unit tests.
2. Run the Example app and check if the icon of fullscreen is being set correctly.
3. On ``ViewController.swift``, append the option ``kFullscreen: true`` and check if the player starts in fullscreen mode.


CristianDRM:
[[Nova UI] Criar botão de fullscreen do MediaControl](https://trello.com/c/oli1OB59/48-nova-ui-criar-bot%C3%A3o-de-fullscreen-do-mediacontrol)

Sources/Clappr_iOS/Classes/Base/FullScreen/FullScreenStateHandler.swift
```diff
@@ -1,48 +1,37 @@
...
 
     var core: Core
 
     func set(fullscreen: Bool) {
-        core.mediaControl?.fullscreen = fullscreen
+        guard core.isFullscreen != fullscreen else { return }
+        core.isFullscreen = fullscreen
+        if fullscreen {
+            core.trigger(InternalEvent.willEnterFullscreen.rawValue)
+            core.trigger(InternalEvent.didEnterFullscreen.rawValue)
```
leandroalonso: I'm having an issue with those events: when the `didEnterFullscreen` is triggered the view is in the middle of the process of going to fullscreen. This should be triggered only after the player is *actually* in fullscreen.

What do you guys think?

CristianDRM:
**@nabilsafatli reports:**

Hey Cristian,

I guess something happened here when the fullscreen mode changes.
35643547-cd802882-06ad-11e8-8426-5d6f4bd18bf9


To repeat this scenario:

With the phone on portrait mode open the app and start with Full screen controlled by app and Start player as fullscreen off.
Press Play
Click on fullscreen
Change the phone to landscape ( in case of a simulator, press Command + <- )
Maybe this is because of the constraints on the Sample app, I am looking into the code to understand what may happened.

CristianDRM:
The issue discovered by @nabilsafatli will be fixed on a different PR.

CristianDRM:
[dsad](https://trello.com/c/Teh0O9TX/66-dsad)

PR #172, opened 2018-02-01
BufferUpdate event should inform buffer duration properly

igoriols:
## Goal

BufferUpdate event was not informing buffer duration properly. We should send the information in the userInfo dictionary with `duration` key.


nabilsafatli:
Can’t we write some tests to avoid this problem on the BufferUpdate event?

PR #174, opened 2018-02-01
Trigger did seek event

igoriols:
## Goal

- We should trigger an event when a seek is completed;
- Fix duration information sent by bufferUpdate event;

## How to test 
1. Play a video;
2. Seek to any point;
3. Check if didSeek event was triggered;

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,2 +1,3 @@
 github "Quick/Quick" == 1.2.0
-github "Quick/Nimble" == 7.0.2
+github "Quick/Nimble" == 7.0.3
+github "AliSoftware/OHHTTPStubs" == 6.1.0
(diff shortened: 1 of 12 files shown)


CristianDRM:
In this context, the seek event is a will seek, right? 
Adding only this event is not a problem for the plugins that are listening to the seek event, but it creates a convention inconsistency in a project, right?

I suggest creating a task to don't forget to analyze this issue, what you think?

**obs: On web and android player doesn't exist did/will seek, only seek, so it's important to document this kind of change.**

nabilsafatli:
Hey @igoriols ,

Don't we need to add the didSeek event on the documentation as well?

igoriols:
Hey guys, the documentation is on the wiki. We will modify the documentation after merging. About renaming the seek event to willSeek. We can discuss about it.

PR #175, opened 2018-02-02
Create willSeek and didSeek events

igoriols:
## Goal
To create two new events:
- willSeek: triggered when user wants to play another part of the video;
- didSeek: triggered when playback will start another part of the video;

**seek event will be deprecated**

## How to test
In iOS and tvOS: 
1. Open any vod;
2. Move scrubber to any point of the video;
3. Check the output if willSeek and didSeek events were triggered;

## TO DO
After merge, update clappr wiki with the new events.

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -239,9 +239,12 @@ open class AVFoundationPlayback: Playback {
         let time = CMTimeMakeWithSeconds(timeInterval, Int32(NSEC_PER_SEC))
 
         trigger(.seek)
+        trigger(.willSeek)
 
         player?.currentItem?.seek(to: time) { [weak self] success in
-            self?.trigger(.didSeek, userInfo: ["success": success])
+            if success {
+                self?.trigger(.didSeek)
+            }
         }
 
         trigger(.positionUpdate, userInfo: ["position": CMTimeGetSeconds(time)])
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -238,8 +238,12 @@ open class AVFoundationPlayback: Playback {
     open override func seek(_ timeInterval: TimeInterval) {
         let time = CMTimeMakeWithSeconds(timeInterval, Int32(NSEC_PER_SEC))
 
-        player?.currentItem?.seek(to: time)
         trigger(.seek)
+
+        player?.currentItem?.seek(to: time) { [weak self] success in
+            self?.trigger(.didSeek, userInfo: ["success": success])
+        }
+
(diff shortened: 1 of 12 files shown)


Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -238,8 +238,15 @@ open class AVFoundationPlayback: Playback {
     open override func seek(_ timeInterval: TimeInterval) {
         let time = CMTimeMakeWithSeconds(timeInterval, Int32(NSEC_PER_SEC))
 
-        player?.currentItem?.seek(to: time)
         trigger(.seek)
+        trigger(.willSeek)
```
CristianDRM: Why did you keep the ``.seek`` event here?

Sources/Clappr_tvOS/Classes/Plugin/Playback/AVFoundationPlayback/AVFoundationPlayback.swift
```diff
@@ -265,13 +265,26 @@ open class AVFoundationPlayback: Playback, AVPlayerViewControllerDelegate {
...
         trigger(.positionUpdate, userInfo: ["position": CMTimeGetSeconds(time)])
     }
 
+    public func playerViewController(_ playerViewController: AVPlayerViewController, timeToSeekAfterUserNavigatedFrom oldTime: CMTime, to targetTime: CMTime) -> CMTime {
+        trigger(.willSeek)
+        return targetTime
+    }
+
     public func playerViewController(_ playerViewController: AVPlayerViewController, willResumePlaybackAfterUserNavigatedFrom oldTime: CMTime, to targetTime: CMTime) {
         trigger(.seek)
```
CristianDRM: on Seek method the old seek event was triggered with willseek, here is triggered with didSeek.
I think that is better to remove the old .seek event from the Project.

igoriols: The previous behavior of seek event was act as a willSeek event. In our metrics plugins the most correct behavior is to listen to didSeek.

We are proposing to keep seek, but deprecate it. We need to update the metrics plugins to listen to the correct event in another task. Our plan is to keep it for now and remove it when we update our metrics plugins.

What you think about that?

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
leandroalonso: We will keep the `seek` event for now and update the docs that it will be deprecated. If we remove it now we can break all of our users that listen to this event.

dezouzart:
Hey guys, at the first time play event is been triggered, will seek and didseek are been triggered too. Is this the expected behaviour?

igoriols:
[[iOS | Clappr] Novos eventos: willSeek e didSeek](https://trello.com/c/xSh4Cyi6/61-ios-clappr-novos-eventos-willseek-e-didseek)

PR #176, opened 2018-02-08
Expose container options

igoriols:
## Goal
Expose container options to be able to change it out of the class.


igoriols:
[[iOS | Clappr] Expondo options do Container](https://trello.com/c/vp2KdIle/62-ios-clappr-expondo-options-do-container)

PR #177, opened 2018-02-08
Set the ui tests to use async approach

CristianDRM:
### Summary
The UI tests in dev branch were broken.

### Goal
Set the UI tests to use async approach.

### How to test
Run the UI tests and check if everything is ok.


dezouzart:
When I ran tests on dev, all have passed. I guess it is intermittent, isn't it? But also, all have passed on your branch. =p

PR #178, opened 2018-02-20
Retrieve all stored cookies

dezouzart:
#### Problem
When we get cookies with different paths, we are not able to retrieve them all from HTTPCookieStorage

#### Goal
Change the way we fetch it

### DIFF
--- a/Sources/Clappr_iOS/Classes/Extension/AVURLAssetWithCookiesBuilder.swift
+++ b/Sources/Clappr_iOS/Classes/Extension/AVURLAssetWithCookiesBuilder.swift
@@ -34,9 +34,7 @@ struct AVURLAssetWithCookiesBuilder: AVURLAssetWithCookies {
     }
 
     private func getCookies() -> [HTTPCookie]? {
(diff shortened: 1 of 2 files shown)


Sources/Clappr_iOS/Classes/Extension/AVURLAssetWithCookiesBuilder.swift
```diff
@@ -34,9 +34,7 @@ struct AVURLAssetWithCookiesBuilder: AVURLAssetWithCookies {
     }
 
     private func getCookies() -> [HTTPCookie]? {
-        if let host = self.url.host, let cookieUrl = URL(string: "http://\(host)") {
-            return HTTPCookieStorage.shared.cookies(for: cookieUrl)
-        }
-        return nil
+        let url = self.url.absoluteString
+        return HTTPCookieStorage.shared.cookies?.filter { url.range(of:$0.domain) != nil }
```
geisonfgf: Since we are only using the domain to filter the cookies my suggestion is to extract the logic to get the domain from the self.url to a computed property.

PR #179, opened 2018-02-22
Set the Loading plugin to be a Core Plugin

CristianDRM:
### Goal
Set the LoadingContainerPlugin to be a plugin of the Core.

### How to test
1. Run all unit and ui tests.
1. Run the app and check if the Spinner appears when the video is buffering and hides when is playing.


Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -1,11 +1,11 @@
 open class Loader {
     internal(set) open var playbackPlugins: [Plugin.Type] = [AVFoundationPlayback.self, NoOpPlayback.self]
     #if os (iOS)
-    internal(set) open var containerPlugins: [Plugin.Type] = [PosterPlugin.self, LoadingContainerPlugin.self]
+    internal(set) open var containerPlugins: [Plugin.Type] = [PosterPlugin.self]
     #else
     internal(set) open var containerPlugins: [Plugin.Type] = []
     #endif
-    internal(set) open var corePlugins = [Plugin.Type]()
+    internal(set) open var corePlugins: [Plugin.Type] = [LoadingCorePlugin.self]
```
igoriols: Hey @CristianDRM , remember to set it only for iOS. In tvOS, we do not need this plugin.

PR #186, opened 2018-03-08
Fix seek 

CristianDRM:
# Goal
To fix seek in iOS 10.3.

# How to test
1. Test the startAt
2. Try to do a seek when the player receives a ready event.
3. Test manually a seek in a video.


nabilsafatli:
Hey guys,

Don't forget to remove the readyToPlay from our events if we won't use it anymore to fix the seek issue.

PR #188, opened 2018-03-09
setup Swifter to stub the requests

CristianDRM:
# Goal
Setup Swifter to stub the requests

# How to test
Run ``make test``


Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -10,6 +11,16 @@ class AVFoundationPlaybackTests: QuickSpec {
     override func spec() {
         describe("AVFoundationPlayback Tests") {
 
+            let server = HTTPStub()
```
dezouzart: Nice! finally we can stub http requests \o/

PR #193, opened 2018-04-23
Update fastlane and xcov

igoriols:
### Goal
To improve compatibility of our development environment with our command line tools, fastlane was updated to version `2.92.1` and xcov was updated to version `1.4.0`. 

### How to update
1. Run `make setup`;
2. Check if fastlane was updated to version `2.92.1`: `bundle exec fastlane --version`;
3. Check if xcov was updated to version `1.4.0`: `bundle exec xcov --version`;

### How to test
Run `make test`.


Gemfile
```diff
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
 
 gem 'cocoapods', '~> 1.3.1'
-gem "fastlane", "~> 2.64"
-gem "xcov", "~> 1.3.0"
+gem "fastlane", "~> 2.92"
```
jchahoud: it seems there is already a 2.93 release, not sure if it is worth it to change it again :)

PR #202, opened 2018-05-06
Auto Hide Home Button on fullscreen mode

nabilsafatli:
### Goal

Hide the Home Button for iPhone X when in Fullscreen mode

### How to test
- Open an iPhone X
- Turn off properties of `Fullscreen controlled by app` and `enter Fullscreen mode`
- Play the Video on Portrait mode
- Press Fullscreen button, after a second you will notice that the home button is hidden automatically with the media control


nabilsafatli:
[[Clappr Issue]](https://trello.com/c/2SzO6DlF/187-clappr-issue)

PR #207, opened 2018-06-19
Destroy player

CristianDRM:
## Goal
When we tried to call the method `destroy()` from a player instance we had memory issues with  `Core`, `FullscreenStateHandler` and `DragDetectorView` that were not deallocated.

Before:
image

After:
image 1

## How to test
Download the [patch](https://gist.github.com/nabilsafatli/cc492a6ba4fe4b65358b51f7ef9d8f9d), and apply it by running the command `git apply {patch file}`

With this [patch](https://gist.github.com/nabilsafatli/cc492a6ba4fe4b65358b51f7ef9d8f9d), you will be able to open Clappr sample, disable the option of `enter fullscreen` and play the video many times creating new instances with the button to make sure that everything is being deallocated.

### DIFF
--- a/Sources/Clappr/Classes/View/DragDetectorView.swift
+++ b/Sources/Clappr/Classes/View/DragDetectorView.swift
@@ -10,7 +10,7 @@ open class DragDetectorView: UIView {
 
     @objc fileprivate(set) open var currentTouch: UITouch?
 
-    @objc open var target: AnyObject?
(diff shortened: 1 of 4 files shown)


nabilsafatli:
[Leak de Memória no Player - Issue #49](https://trello.com/c/LEILjqMV/233-leak-de-mem%C3%B3ria-no-player-issue-49)

PR #209, opened 2018-07-03
DVR Integration and Events

nabilsafatli:
### Goal

Implement two events to make sure that DVR works in the player and other plugins have what they need of information to change when necessary

- `supportDVR` The current live media being played, supports DVR
- `usingDVR` The current live media being played is using the DVR ( it means that the scrubber is not in the end of the scrubber)


nabilsafatli:
[[DVR] Notificar quando mídia está em estado de DVR](https://trello.com/c/6YLhZz4K/266-dvr-notificar-quando-m%C3%ADdia-est%C3%A1-em-estado-de-dvr)

PR #210, opened 2018-07-17
Adjust DVR logic to follow architecture 

igoriols:
### Goal

Remove `DVRPlugin` and add all dvr logic into `AVFoundationPlayback`.

### Additional Info
1. To check if dvr is available, we created a new property (`supportDVR`) in `AVFoundationPlayback`;
2. When user seeks to a moment and the video has dvr enabled, `AVFoundationPlayback` will trigger `Event.usingDVR` with a flag `enabled` in userInfo. If it is in dvr. The flag will indicate if it is using dvr (true) or not (false);

### How to test
Run `make test`.


igoriols:
[[DVR] Enviar evento de vídeo em modo DVR](https://trello.com/c/b0LGRiLq/271-dvr-enviar-evento-de-v%C3%ADdeo-em-modo-dvr)

Sources/Clappr_iOS/Classes/Base/Playback.swift
```diff
@@ -120,3 +120,26 @@ open class Playback: UIBaseObject, Plugin {
...
+
+    @objc var usingDVR: Bool {
+        return false
+    }
+
+    @objc open var seekableTimeRanges: [NSValue] {
+        return []
+    }
+
+    @objc open var loadedTimeRanges: [NSValue] {
```
CristianDRM: why do we need to expose this property in Playback class?

Sources/Clappr_iOS/Classes/Base/Playback.swift
```diff
@@ -120,3 +120,26 @@ open class Playback: UIBaseObject, Plugin {
...
+extension Playback {
+    @objc var minDvrSize: Double {
+        return 0
+    }
+
+    @objc var usingDVR: Bool {
+        return false
+    }
+
+    @objc open var seekableTimeRanges: [NSValue] {
```
CristianDRM: Why do we need to expose the seekableTimeRange in this way? Could we be more specific? Like a ClosedRange for example.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -265,19 +268,29 @@ open class AVFoundationPlayback: Playback {
     }
 
     open override func seek(_ timeInterval: TimeInterval) {
+        var timeToSeek = timeInterval
+
+        if supportDVR, let seekStart = seekableTimeRanges.first?.timeRangeValue.start.seconds {
```
CristianDRM: Why do we need to always get the first timeRangeValue? There is any chance to the second be the right one?

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -518,9 +518,16 @@ extension AVFoundationPlayback {
     }
 
     open override var usingDVR: Bool {
+        guard playbackType == .live else { return false }
         guard let currentTime = player?.currentTime().seconds else { return false }
-        guard let seekableEndTime = seekableTimeRanges.first?.timeRangeValue.end.seconds else { return false }
-        return playbackType == .live && currentTime < seekableEndTime
+
+        for range in seekableTimeRanges {
```
brunovpl: I think we can inline this _for_ using something like this:

`return seekableTimeRanges.first(where: { $0.timeRangeValue.end.seconds > currentTime})  != nil`

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -534,7 +541,14 @@ extension AVFoundationPlayback {
     }
 
     open override var supportDVR: Bool {
-        guard let duration = seekableTimeRanges.first?.timeRangeValue.duration.seconds else { return false }
-        return playbackType == .live && duration >= minDvrSize
+        guard playbackType == .live else { return false }
+
+        for range in seekableTimeRanges {
```
brunovpl: I think we can inline this _for_ using something like this:

`return seekableTimeRanges.first(where: { $0.timeRangeValue. duration.seconds > minDvrSize})  != nil`

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -489,3 +510,45 @@ open class AVFoundationPlayback: Playback {
...
+extension AVFoundationPlayback {
+    open override var minDvrSize: Double {
+        return 60.0
+    }
+
+    open override var usingDVR: Bool {
+        guard playbackType == .live else { return false }
+        guard let currentTime = player?.currentTime().seconds else { return false }
+
+        for range in seekableTimeRanges {
```
brunovpl: I think we can inline this _for_ using something like this:

`return seekableTimeRanges.first(where: { $0.timeRangeValue.end.seconds > currentTime})  != nil`

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -489,3 +510,45 @@ open class AVFoundationPlayback: Playback {
...
+
+    open override var loadedTimeRanges: [NSValue] {
+        guard let ranges = player?.currentItem?.loadedTimeRanges else { return [] }
+        return ranges
+    }
+
+    open override var supportDVR: Bool {
+        guard playbackType == .live else { return false }
+
+        for range in seekableTimeRanges {
```
brunovpl: I think we can inline this _for_ using something like this:

`return seekableTimeRanges.first(where: { $0.timeRangeValue. duration.seconds > minDvrSize})  != nil`

igoriols: Done!

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
igoriols: Done!

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
igoriols: Hey @CristianDRM, I changed the code to select the correct time range. Please, take a look.

Sources/Clappr_iOS/Classes/Base/Playback.swift
igoriols: Hey @CristianDRM, we added those properties here in order to not cast `Playback` to `AVFoundationPlayback`. If a new playback is created, those properties will be overridden when needed.

PR #211, opened 2018-07-18
Add DVR minimum size to Options

geisonfgf:
### Goal
Enable minDvrTime be set by options

### How to test
1 - In Example app set the option[kMinDvrOption] to 15
2 - Run the app and check if is possible make a seek in a Live video
3 - Run all tests

### DIFF
--- a/Sources/Clappr_iOS/Classes/Base/Options.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Options.swift
@@ -11,7 +11,7 @@ public let kPlaybackNotSupportedMessage = "playbackNotSupportedMessage"
 public let kMimeType = "mimeType"
 public let kDefaultSubtitle = "defaultSubtitle"
 public let kDefaultAudioSource = "defaultAudioSource"
-public let kMinDvrOption = "minDvrTime"
+public let kMinDvrSize = "minDvrSize"
 
 struct OptionsUnboxer {
     let options: Options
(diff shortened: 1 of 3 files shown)


Sources/Clappr_iOS/Classes/Base/Options.swift
```diff
@@ -11,6 +11,7 @@ public let kPlaybackNotSupportedMessage = "playbackNotSupportedMessage"
 public let kMimeType = "mimeType"
 public let kDefaultSubtitle = "defaultSubtitle"
 public let kDefaultAudioSource = "defaultAudioSource"
+public let kMinDvrOption = "minDvrTime"
```
igoriols: Hey @geisonfgf , I suggest change the name to `kMinDvrTime` to make it self explaining.

dezouzart: web is using this name `minDvrSize`

dezouzart:
@geisonfgf , should we update the docs to include this new option?

geisonfgf:
@dezouzart I had understood that this option is only for tests so it should not be available in the docs.

Sources/Clappr_iOS/Classes/Base/Options.swift
nabilsafatli: Hey guys, let's use the Player Web name.

nabilsafatli:
Hey guys,
The PR name is `feat: enable minDvrTime be set by options`

nabilsafatli:
Hey @geisonfgf ,
Each client player has its needs and may expect a different minimum size for DVR.
Even though it was for testing reasons, as it is available here it should be documented. 😄

igoriols:
@nabilsafatli , we added `kMinDvrSize` to the [options list](https://github.com/clappr/clappr-ios/wiki/Options).

PR #212, opened 2018-07-18
Adds a property to return the position inside a DVR window

leandroalonso:
# Goal

Adds a property that returns the position inside a DVR window. Let's say we have a window of 60 seconds, starting at the second 120 and ending in 180.

If the position is `150` the `dvrPosition` should return `30` (a value between `0` and the duration of the window)

### DIFF
--- a/Sources/Clappr_iOS/Classes/Base/Playback.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Playback.swift
@@ -142,4 +142,8 @@ extension Playback {
     @objc open var supportDVR: Bool {
         return false
     }
+
+    @objc open var dvrPosition: Double {
+        return 0
+    }
 }
--- a/Sources/Clappr_iOS/Classes/Base/Playback.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Playback.swift
@@ -127,7 +127,7 @@ extension Playback {
         return 0
     }
 
-    @objc var usingDVR: Bool {
+    @objc open var usingDVR: Bool {
         return false
     }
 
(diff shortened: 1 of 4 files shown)


leandroalonso:
[[DVR] Scrubber](https://trello.com/c/ukKU2xTW/256-dvr-scrubber)

Sources/Clappr_iOS/Classes/Base/Playback.swift
```diff
@@ -127,7 +127,7 @@ extension Playback {
         return 0
     }
 
-    @objc var usingDVR: Bool {
+    @objc open var usingDVR: Bool {
```
igoriols: Hey @leandroalonso and @dezouzart , are you using this property in another module? If the event is triggered only when a change in dvr status is observed, there is no need to pass `usingDVR` in userInfo. I suggest change the name of the event to something like `didChangeDvrStatus`.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -565,4 +565,16 @@ extension AVFoundationPlayback {
         guard playbackType == .live else { return false }
         return seekableTimeRanges.first(where: { $0.timeRangeValue.duration.seconds >= minDvrSize}) != nil
     }
+
+    open override var dvrPosition: Double {
+        if let start = seekableTimeRanges.first?.timeRangeValue.start.seconds,
```
geisonfgf: Instead of use seekableTimeRanges.first? I think we need check all seekableTimeRanges and get the lower start value in order to guarantee the start seekable position.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -565,4 +565,16 @@ extension AVFoundationPlayback {
...
+            let end = seekableTimeRanges.first?.timeRangeValue.end.seconds,
```
geisonfgf: Instead of use seekableTimeRanges.first? I think we need check all seekableTimeRanges and get the higher end value in order to guarantee the end seekable position.

Sources/Clappr_iOS/Classes/Base/Playback.swift
igoriols: Hey guys, I'm going to approve this PR. As we discussed before, those changes will be made in another PR.

PR #215, opened 2018-07-25
Expose property to inform media current date

igoriols:
### Goal

To get the thumbseek image for a live video with dvr, it is necessary to obtain the date of the media. So, we need to expose a property informing the media current date. 

### Additional Information

This property is loaded from master playlist tag `#EXT-X-PROGRAM-DATE-TIME`. 

### How to test

Run `make test`.

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -574,7 +574,7 @@ extension AVFoundationPlayback {
         return position
     }
 
-    open var currentDate: Date? {
+    open override var currentDate: Date? {
         return player?.currentItem?.currentDate()
     }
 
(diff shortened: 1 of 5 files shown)


Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -574,6 +574,10 @@ extension AVFoundationPlayback {
         return position
     }
 
+    open var currentDate: Date? {
```
nabilsafatli: Hey guys,

If we need this information available for other `Plugins` then it should be added in the [Playback.swift](https://github.com/clappr/clappr-ios/blob/dev/Sources/Clappr_iOS/Classes/Base/Playback.swift) class.

The access for any information should never cast the Playback to a specific implementation like:
❎ Indicating that we are using not following our architecture:
```
let playback = self.currentPlayback as? AvFoundationPlayback
let date = playback?.currentDate
```

igoriols: Done!

PR #216, opened 2018-07-25
Refactoring DVR to keep Clappr architecture consistency

nabilsafatli:
### Goal

After talking with the Clappr team, some names were changed to inform better the DVR availability and status.

- Create an event to indicate the change of DVR availability - didChangeDvrAvailability
- Change the property `supportDvr` to `isDvrAvailable`
- Change the property `isUsingDvr` to `isDvrInUse`
- Change the event `usingDvr` to `didChangeDvrStatus`
- Create a method `seekToLivePosition()` to make possible for plugins to redirect a live playback to last live moment available

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -131,6 +131,15 @@ open class AVFoundationPlayback: Playback {
     }
 
     open override var position: Double {
+        if let start = dvrWindowStart,
+            let end = dvrWindowEnd,
+            let position = player?.currentItem?.currentTime().seconds {
+            var calculatedPosition = (position - start) * 100
+            calculatedPosition = calculatedPosition / ((end - start) / 100)
+            calculatedPosition = (calculatedPosition * duration) / 10000
+            return calculatedPosition
+        }
+        
         guard playbackType == .vod, let player = self.player else {
             return 0
         }
@@ -572,6 +581,7 @@ extension AVFoundationPlayback {
     open override var currentDate: Date? {
         return player?.currentItem?.currentDate()
     }
+
     open override var seekableTimeRanges: [NSValue] {
         guard let ranges = player?.currentItem?.seekableTimeRanges else { return [] }
         return ranges
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -593,10 +593,16 @@ extension AVFoundationPlayback {
     }
 
     private var dvrWindowStart: Double? {
-        return seekableTimeRanges.min { rangeA, rangeB in rangeA.timeRangeValue.start.seconds < rangeB.timeRangeValue.start.seconds }?.timeRangeValue.start.seconds
+        guard let end = dvrWindowEnd, isDvrAvailable, playbackType == .live else {
+            return nil
+        }
+        return end - duration
     }
 
     private var dvrWindowEnd: Double? {
+        guard isDvrAvailable, playbackType == .live else {
+            return nil
+        }
         return seekableTimeRanges.max { rangeA, rangeB in rangeA.timeRangeValue.end.seconds < rangeB.timeRangeValue.end.seconds }?.timeRangeValue.end.seconds
     }
 }
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -568,9 +568,9 @@ extension AVFoundationPlayback {
     }
 
     open override var isDvrInUse: Bool {
-        guard playbackType == .live else { return false }
+        guard let end = dvrWindowEnd, playbackType == .live else { return false }
         guard let currentTime = player?.currentTime().seconds else { return false }
-        return seekableTimeRanges.first(where: { $0.timeRangeValue.end.seconds > currentTime}) != nil
+        return end > currentTime
     }
 
     open override var isDvrAvailable: Bool {
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -120,11 +120,11 @@ open class AVFoundationPlayback: Playback {
             return CMTimeGetSeconds(item.asset.duration)
         }
         
-        if playbackType == .live, isDvrAvailable {
-            var liveDuration: Double = 0
-            for seekableTimeRange in seekableTimeRanges {
-                liveDuration += seekableTimeRange.timeRangeValue.duration.seconds
-            }
+        if playbackType == .live {
+            let liveDuration = seekableTimeRanges.reduce(0.0, { previous, current in
+                return previous + current.timeRangeValue.duration.seconds
+            })
+
             return liveDuration
         }
 
@@ -572,13 +572,8 @@ extension AVFoundationPlayback {
 
     open override var isDvrAvailable: Bool {
         guard playbackType == .live else { return false }
-        var liveDuration: Double = 0
         
-        for seekableTimeRange in seekableTimeRanges {
-            liveDuration += seekableTimeRange.timeRangeValue.duration.seconds
-        }
-        
-        return liveDuration >= minDvrSize
+        return duration >= minDvrSize
     }
 
     open override var currentDate: Date? {
--- a/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -120,7 +120,7 @@ open class AVFoundationPlayback: Playback {
         }
 
         if playbackType == .live {
-            if supportDVR, let duration = seekableTimeRanges.first?.timeRangeValue.duration.seconds {
+            if isDvrAvailable, let duration = seekableTimeRanges.first?.timeRangeValue.duration.seconds {
                 return duration
             } else {
                 return 0
@@ -281,7 +281,7 @@ open class AVFoundationPlayback: Playback {
     }
 
     open override func seek(_ timeInterval: TimeInterval) {
-        if supportDVR {
+        if isDvrAvailable {
             var timeToSeek = timeInterval
 
             if let seekStartTime = dvrWindowStart {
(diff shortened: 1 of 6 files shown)


nabilsafatli:
[[DVR] Normalizar nomes entre as plataformas](https://trello.com/c/uhlMRQyp/280-dvr-normalizar-nomes-entre-as-plataformas)

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -38,6 +38,7 @@ open class AVFoundationPlayback: Playback {
     }
     fileprivate var timeObserver: Any?
     fileprivate var asset: AVURLAsset?
+    private var lastDvrAvailability: Bool?
```
nabilsafatli: @dezouzart , can you help me by creating a test to cover a scenario that use this variable?
Example: `Only triggers didChangeDvrAvailability when lastDvrAvailability is different`

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -131,6 +132,15 @@ open class AVFoundationPlayback: Playback {
     }
 
     open override var position: Double {
+        if let start = dvrWindowStart,
+            let end = dvrWindowEnd,
+            let position = player?.currentItem?.currentTime().seconds {
+            var calculatedPosition = (position - start) * 100
```
towerz: Is there any reason to have this calculation here? It seems to be unnecessary.

Assuming that `duration == end - start` (at least it should be. The duration reported by the playback should be the duration of the DVR seekable window):
```
                      (position - start) * 100     duration
calculatedPosition = -------------------------- * -----------
                        (end - start) / 100          10000
                      (position - start) * 10000     duration
calculatedPosition = ---------------------------- * ----------
                           (end - start)              10000
                         (position - start)
calculatedPosition = -------------------------- * duration
                           (end - start)
                         (position - start)
calculatedPosition = -------------------------- * duration
                              duration

calculatedPosition = position - start            
```

towerz: Speaking of the `dvrWindow{Start|End}`, we need to take into account that the seekableTimeRanges could be discontinuous. So merely subtracting could be an issue if there's any discontinuity in the timeline.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -541,12 +564,21 @@ extension AVFoundationPlayback {
...
-    open override var usingDVR: Bool {
+    open override var isDvrInUse: Bool {
         guard playbackType == .live else { return false }
         guard let currentTime = player?.currentTime().seconds else { return false }
         return seekableTimeRanges.first(where: { $0.timeRangeValue.end.seconds > currentTime}) != nil
     }
 
+    open override var isDvrAvailable: Bool {
+        guard playbackType == .live else { return false }
+        return seekableTimeRanges.first(where: { $0.timeRangeValue.duration.seconds >= minDvrSize}) != nil
```
towerz: Why are we considering a single seekableTimeRange? The total duration should be taken into account, not a single range.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -557,27 +589,6 @@ extension AVFoundationPlayback {
...
-        }
-        return position
-    }
-
-    open override var currentDate: Date? {
-        return player?.currentItem?.currentDate()
-    }
-
     private var dvrWindowStart: Double? {
         return seekableTimeRanges.min { rangeA, rangeB in rangeA.timeRangeValue.start.seconds < rangeB.timeRangeValue.start.seconds }?.timeRangeValue.start.seconds
```
towerz: I think it's risky to calculate the boundaries between seekableTimeRanges. The time ranges could be discontinuous, so the difference between the start of the first range and the end of the last range may not represent the duration.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -120,7 +121,7 @@ open class AVFoundationPlayback: Playback {
         }
 
         if playbackType == .live {
-            if supportDVR, let duration = seekableTimeRanges.first?.timeRangeValue.duration.seconds {
+            if isDvrAvailable, let duration = seekableTimeRanges.first?.timeRangeValue.duration.seconds {
```
towerz: About the duration, we should be careful about considering the duration of only the first seekable range. If there's a discontinuity between ranges, we are discarding the other ranges from this duration, so it would report a smaller value than it should be. The right way of reporting this is to use the duration of all seekable time ranges.

nabilsafatli: Thanks for the feedback Towerz. Is it possible to get a playlist with this discontinuity scenario to see how the seekableTimeRanges changes?

In all our testing scenarios, the first item of the array was the only one returning a value, that’s why we kept only one Time Range.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
nabilsafatli: As I wrote in the previous reply, we will try to reproduce this scenario to test what values return and make sure that the logic works, otherwise find the right solution.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
towerz: @nabilsafatli sure, I'm arranging to get a live video simulating this kind of behavior. As soon as I have this ready I'll let you know.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -38,6 +38,7 @@ open class AVFoundationPlayback: Playback {
     }
     fileprivate var timeObserver: Any?
     fileprivate var asset: AVURLAsset?
+    var lastDvrAvailability: Bool?
```
igoriols: Hey @nabilsafatli  and @dezouzart , is it necessary to be optional?

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -281,20 +287,22 @@ open class AVFoundationPlayback: Playback {
...
+    }
 
-            seek(timeToSeek) { [weak self] in
-                if let usingDVR = self?.usingDVR {
-                    self?.trigger(.usingDVR, userInfo: ["enabled": usingDVR])
-                }
-            }
+    private func relativeTime(to time: TimeInterval) -> TimeInterval {
+        if isDvrAvailable {
+            return time + (dvrWindowStart ?? 0)
```
towerz: We need to check these relative time calculations when there are discontinuities, just to have a predictable behavior.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -541,10 +562,26 @@ extension AVFoundationPlayback {
...
     }
 
-    open override var usingDVR: Bool {
-        guard playbackType == .live else { return false }
+    open override var isDvrInUse: Bool {
+        if isPaused && isDvrAvailable { return true }
+        guard let end = dvrWindowEnd, playbackType == .live else { return false }
         guard let currentTime = player?.currentTime().seconds else { return false }
-        return seekableTimeRanges.first(where: { $0.timeRangeValue.end.seconds > currentTime}) != nil
+        return end > currentTime
```
towerz: Same thing here, this may not work with a discontinuity, since the `currentTime` can be in a different time scale.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -541,10 +562,26 @@ extension AVFoundationPlayback {
...
+
+    open override var isDvrAvailable: Bool {
+        guard playbackType == .live else { return false }
+        var liveDuration: Double = 0
+        
+        for seekableTimeRange in seekableTimeRanges {
+            liveDuration += seekableTimeRange.timeRangeValue.duration.seconds
+        }
+        
+        return liveDuration >= minDvrSize
```
towerz: Instead of repeating the same calculation done in `duration` we can use the `duration` value here.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -321,6 +329,11 @@ open class AVFoundationPlayback: Playback {
         trigger(.positionUpdate, userInfo: ["position": CMTimeGetSeconds(time)])
     }
 
+    open override func seekToLivePosition() {
+        play()
```
igoriols: Hey @nabilsafatli and @dezouzart , why do need to always call `play` here? This will always trigger events related with play. Would we check if it is paused to call play?

towerz: @igoriols seeking to live should resume the playback. Otherwise, it would keep in DVR, only paused at the live position at the moment of the seek.

nabilsafatli: Hey @igoriols ,
When the user ask to seek to live position we assume that the player needs to play again. Otherwise the player will come to a dvr state again.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -566,7 +567,7 @@ extension AVFoundationPlayback {
         if isPaused && isDvrAvailable { return true }
         guard let end = dvrWindowEnd, playbackType == .live else { return false }
         guard let currentTime = player?.currentTime().seconds else { return false }
-        return end > currentTime
+        return end - liveHeadTolerance > currentTime
```
igoriols: Hey @nabilsafatli and @dezouzart , I printed the value of `isDvrInUse` when a positionUpdate is received and even subtracting `liveHeadTolerance`, the property reported that a video was in dvr when the scrubber is on the end of the seekbar. This problem happens when `currentTime` is near `end`.

igoriols: As we talked before, this problem is analyzed in another task.

PR #218, opened 2018-08-20
Configure

leandroalonso:
# Goal

To allow the player options to be changed in the same player instance.

# How to test

1. In `ViewController.swift` add those lines at the end of `viewDidLoad`:

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 10) { [weak self] in
            self?.player.configure(options: [
                kSourceUrl: "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4",
                kPosterUrl: "https://upload.wikimedia.org/wikipedia/commons/thumb/c/c0/Big_Buck_Bunny_4K.webm/310px-seek%3D116-Big_Buck_Bunny_4K.webm.jpg",
                kMediaControl: true
            ])
        }
```
2. Run the example, click in `Start video`
3. After 10s the poster should change, when you play the video the big buck bunny should play
4. If you want the video to auto-start playing you should add `play` after configuring:
```swift
        DispatchQueue.main.asyncAfter(deadline: .now() + 10) { [weak self] in
            self?.player.configure(options: [
                kSourceUrl: "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4",
                kPosterUrl: "https://upload.wikimedia.org/wikipedia/commons/thumb/c/c0/Big_Buck_Bunny_4K.webm/310px-seek%3D116-Big_Buck_Bunny_4K.webm.jpg",
                kMediaControl: true
            ])
            self?.player.play()
        }
```

## Checking if it is the same instance

The method `configure` should not destroy and instantiate the player components. To see if the same `Core` instance is kept, check the logs for something like that:

```
INFO [Clappr] loading with ["posterUrl": "http://clappr.io/poster.png", "fullscreen": true, "fullscreenByApp": true, "sourceUrl": "http://clappr.io/highline.mp4"]
```

This log should only appear once per player instance.

# Docs

The documentation is already changed to reflect this new method: https://github.com/clappr/clappr-ios/wiki/Options#configure


BruunoMM:
I've tested the project on an iPhone X, and I've noticed an odd behaviour, don't know if it should happen or not.

**What I did:**
I added the suggested test code, calling the play function so it'd play as soon as it loads. The video started after 10 seconds and I sought to somewhere near the end of the video.
The video ends and then the load spinner shows up, at this point I'm not able to play the video again and the media control doesn't show up.

I noticed that behaviour at the end of the video, apparently when the machine is on the "On Stalled" state.

PR #219, opened 2018-08-21
Create a way to share data between plugins

dezouzart:
#### Goal
Create a share structure on container that all plugins can store and retrieve data

#### How to test
* Run tests
* Use sharedData in 2 different plugins

### DIFF
--- a/Sources/Clappr/Classes/Base/SharedData.swift
+++ b/Sources/Clappr/Classes/Base/SharedData.swift
@@ -1,0 +1,6 @@
+import Foundation
+
+@objc open class SharedData: NSObject {
+    @objc public var storeDictionary = [String: Any]()
+    @objc public weak var container: Container?
+}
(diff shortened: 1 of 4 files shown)


Sources/Clappr/Classes/Base/SharedData.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+@objc open class SharedData: NSObject {
+    @objc public var storeDictionary = [String: Any]()
+    @objc public weak var container: Container?
```
igoriols: Hey @dezouzart, why `SharedData` has a weak reference to container?

igoriols: Hey @dezouzart , I'm going to aprove this PR. In the future, we can think in another approach to not need this weak reference.

PR #221, opened 2018-09-12
Adjust audio and subtitle default selection

dezouzart:
### Goal
To trigger audioAvailable and subtitleAvailable events informing if has default value

### How to test
Run tests

### DIFF
--- a/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
+++ b/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
@@ -1,8 +1,8 @@
 import AVFoundation
 
 open class AvailableMediaOptions {
-    let sources: [MediaOption]
-    let hasDefaultSelected: Bool
+    open let sources: [MediaOption]
+    open let hasDefaultSelected: Bool
 
     init(_ sources: [MediaOption], hasDefaultSelected: Bool) {
         self.sources = sources
--- a/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
+++ b/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
@@ -1,6 +1,6 @@
-open class AvailableMediaOptions {
-    open let sources: [MediaOption]
-    open let hasDefaultSelected: Bool
+public struct AvailableMediaOptions {
+    public let sources: [MediaOption]
+    public let hasDefaultSelected: Bool
 
     public init(_ sources: [MediaOption], hasDefaultSelected: Bool) {
         self.sources = sources
--- a/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
+++ b/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
@@ -1,0 +1,11 @@
+import AVFoundation
+
+open class AvailableMediaOptions {
+    let sources: [MediaOption]
+    let hasDefaultSelected: Bool
+
+    init(_ sources: [MediaOption], hasDefaultSelected: Bool) {
+        self.sources = sources
+        self.hasDefaultSelected = hasDefaultSelected
+    }
+}
--- a/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
+++ b/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
@@ -1,5 +1,3 @@
-import AVFoundation
-
 open class AvailableMediaOptions {
     open let sources: [MediaOption]
     open let hasDefaultSelected: Bool
--- a/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
+++ b/Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
@@ -4,7 +4,7 @@ open class AvailableMediaOptions {
     open let sources: [MediaOption]
     open let hasDefaultSelected: Bool
 
-    init(_ sources: [MediaOption], hasDefaultSelected: Bool) {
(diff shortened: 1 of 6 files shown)


Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -450,16 +450,36 @@ open class AVFoundationPlayback: Playback {
...
         if let audioSources = self.audioSources {
-            trigger(.audioAvailable, userInfo: ["audios": audioSources])
+            let hasDefault = selectDefaultAudioIfNeeded()
+            trigger(.audioAvailable, userInfo: ["audios": audioSources, "hasDefaultFromOption": hasDefault])
         }
 
         addTimeElapsedCallback()
     }
 
+    fileprivate func selectDefaultSubtitleIfNeeded() -> Bool {
```
paulogama14: Why do we need to return a Bool value?  I think that we have more than one responsibility in this method.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -820,6 +820,80 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                            let playback = AVFoundationPlayback(options: options)
+                            var hasDefaultFromOption = false
+                            playback.on(Event.subtitleAvailable.rawValue) { (userInfo: EventUserInfo) in
+                                guard let hasDefault = userInfo?["hasDefaultFromOption"] as? Bool else { return }
+                                hasDefaultFromOption = hasDefault
+                            }
+                            
+                            playback.play()
+                            
+                            expect(hasDefaultFromOption).toEventually(beTrue())
```
leandroalonso: This test is failing here. Increasing the timeout to `2` makes it pass.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -820,6 +820,80 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                            let playback = AVFoundationPlayback(options: options)
+                            var hasDefaultFromOption = true
+                            playback.on(Event.subtitleAvailable.rawValue) { (userInfo: EventUserInfo) in
+                                guard let hasDefault = userInfo?["hasDefaultFromOption"] as? Bool else { return }
+                                hasDefaultFromOption = hasDefault
+                            }
+                            
+                            playback.play()
+                            
+                            expect(hasDefaultFromOption).toEventually(beFalse())
```
leandroalonso: Same as before.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -450,16 +450,36 @@ open class AVFoundationPlayback: Playback {
         seekOnReadyIfNeeded()
 
         if let subtitles = self.subtitles {
-            trigger(.subtitleAvailable, userInfo: ["subtitles": subtitles])
+            let hasDefault = selectDefaultSubtitleIfNeeded()
+            trigger(.subtitleAvailable, userInfo: ["subtitles": subtitles, "hasDefaultFromOption": hasDefault])
```
leandroalonso: `hasDefaultFromOption` is true if the user passed any audio/subtitle preference and it was selected, correct?

Maybe we can change this to `defaultSubtitleSelected`.

Sources/Clappr_iOS/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -450,16 +450,36 @@ open class AVFoundationPlayback: Playback {
...
         }
 
         if let audioSources = self.audioSources {
-            trigger(.audioAvailable, userInfo: ["audios": audioSources])
+            let hasDefault = selectDefaultAudioIfNeeded()
+            trigger(.audioAvailable, userInfo: ["audios": audioSources, "hasDefaultFromOption": hasDefault])
```
leandroalonso: Same as before. We can name `hasDefaultFromOption` to `defaultAudioSelected`.

Sources/Clappr/Classes/Base/AvailableMediaOptions.swift
```diff
@@ -0,0 +1,11 @@
+import AVFoundation
```
igoriols: No need to import `AVFoundation` here.

Sources/Clappr_tvOS/Classes/Plugin/Playback/AVFoundationPlayback/AVFoundationPlayback.swift
```diff
@@ -455,6 +474,7 @@ open class AVFoundationPlayback: Playback {
         if keyPath == "currentItem.playbackLikelyToKeepUp" {
             if player?.currentItem?.isPlaybackLikelyToKeepUp == true && currentState == .buffering {
                 play()
+                selectDefaultSubtitleIfNeeded()
```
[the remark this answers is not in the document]
dezouzart: AVPlayer does not support 2 different media option set on earlier moment. You can check in `Select the Desired Media Option` session in [apple doc](https://developer.apple.com/documentation/avfoundation/media_assets_playback_and_editing/adding_subtitles_and_alternative_audio_tracks).
So I decide to set audio as soon as possible and subtitles later.

PR #223, opened 2018-09-20
Remove unnecessary events rethrow

brunovpl:
### Goal
After the study to find duplicated triggered events at our code. We found some unnecessary events been throw by our MediaPlayer. To fix that and be sure that our users do not listen events from the wrong place, we decided to remove these unnecessary event triggers.

### How to test
1. Run clappr tests


dezouzart:
hey @brunovpl could you give me more info/context? I mean, why don't we need to rethrow playback events anymore? or why did we rethrow it until now?

PR #225, opened 2018-09-28
Remove duplicate Ready event in tvOS

igoriols:
### Goal
- Remove duplicate Ready event in tvOS;
- Unifying NoOpPlayback for iOS and tvOS;

### How to test
Run `make test`.


dezouzart:
@igoriols , there were a difference between ios and tvos NoOpPlayback's render method.
`addSubviewMatchingConstraints` were commented on tvos. Do you know why? and/or should we call it on tvos without problem?

igoriols:
Hey @dezouzart, it was commented to not render the label with "Could not play video". It was commented in the beginning of tvOS development. We solved it using the option `option[kPlaybackNotSupportedMessage] = "" as AnyObject?` to not display the message. So it is safe to leave it like that.

PR #226, opened 2018-09-28
Refactoring Loader

nabilsafatli:
### Goal

The objective is to change `Loaders` class responsibility in the architecture.
Instead of holding the `MediaControl` and `Options` reference, the `Loader` now is responsible for managing and holding plugins reference across the Player lifecycle.

[ ------- **Loader** ------- ]
| _containerPlugins_ 
| _corePlugins_
| _playbacks_ **changed the name from playbackPlugins to playback
| _mediaControl_ **removed
| _externalPlugins_
[---------------------------]

The `Loader` is now a Singleton and won't receive the external plugins and options in its constructor.

### How to test

- Open the player and check if all controls are available correctly without any problem
- Run all tests and make sure they pass

### DIFF
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -9,8 +9,6 @@ open class Container: UIBaseObject {
         }
     }
 
-    fileprivate var loader: Loader
-
     @objc open var mediaControlEnabled = false {
         didSet {
             let eventToTrigger: Event = mediaControlEnabled ? .enableMediaControl : .disableMediaControl
@@ -33,10 +31,9 @@ open class Container: UIBaseObject {
         }
     }
 
-    public init(loader: Loader = Loader(), options: Options = [:]) {
+    public init(options: Options = [:]) {
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
         self.options = options
-        self.loader = loader
         super.init(frame: CGRect.zero)
         self.sharedData.container = self
         backgroundColor = UIColor.clear
@@ -60,7 +57,7 @@ open class Container: UIBaseObject {
 
         self.playback?.destroy()
 
-        let playbackFactory = PlaybackFactory(loader: loader, options: playbackOptions)
+        let playbackFactory = PlaybackFactory(options: playbackOptions)
         self.playback = playbackFactory.createPlayback()
 
         if playback is NoOpPlayback {
@@ -93,7 +90,7 @@ open class Container: UIBaseObject {
     }
 
     fileprivate func loadPlugins() {
-        for type in loader.containerPlugins {
+        for type in Loader.shared.containerPlugins {
             if let plugin = type.init(context: self) as? UIContainerPlugin {
                 addPlugin(plugin)
             }
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -37,7 +37,7 @@ open class Container: UIBaseObject {
         super.init(frame: CGRect.zero)
         self.sharedData.container = self
         backgroundColor = UIColor.clear
-        loadPlugins()
+        Loader.shared.loadPlugins(in: self)
         accessibilityIdentifier = "Container"
         if let source = options[kSourceUrl] as? String {
             load(source, mimeType: options[kMimeType] as? String)
@@ -89,15 +89,7 @@ open class Container: UIBaseObject {
         plugin.render()
     }
 
-    fileprivate func loadPlugins() {
-        for type in Loader.shared.containerPlugins {
-            if let plugin = type.init(context: self) as? UIContainerPlugin {
-                addPlugin(plugin)
-            }
-        }
-    }
-
-    private func addPlugin(_ plugin: UIContainerPlugin) {
+    func addPlugin(_ plugin: UIContainerPlugin) {
         plugins.append(plugin)
     }
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -75,7 +75,7 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         #endif
 
         bindEventListeners()
-        loadPlugins()
+        Loader.shared.loadPlugins(in: self)
 
         containers.append(Container(options: options))
 
@@ -88,14 +88,6 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     fileprivate func setActive(container: Container) {
         if activeContainer != container {
             activeContainer = container
-        }
-    }
-
-    fileprivate func loadPlugins() {
-        for plugin in Loader.shared.corePlugins {
-            if let corePlugin = plugin.init(context: self) as? UICorePlugin {
-                addPlugin(corePlugin)
-            }
         }
     }
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -54,7 +54,7 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         fatalError("Should be using init(sources:[NSURL]) instead")
     }
 
-    public required init(loader: Loader = Loader(), options: Options = [:]) {
+    public required init(options: Options = [:]) {
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
 
         self.options = options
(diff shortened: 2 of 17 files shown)


Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -1,63 +1,46 @@
...
-        self.externalPlugins = externalPlugins
-
-        loadExternalMediaControl(options)
-
-        if !externalPlugins.isEmpty {
-            addExternalPlugins(externalPlugins)
-        }
-
+    
+    var externalPlugins: [Plugin.Type] = []
```
towerz: Just to reiterate, let's not differentiate between registered and external plugins (there shouldn't be any hard-coded, built-in plugins).

The external plugins (passed via options) should only be used when loading the plugins into a context. The Loader instance shouldn't store those. This could cause plugins to leak from a Player instance to another.

Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -1,63 +1,46 @@
...
-    fileprivate func getPlugins(_ type: PluginType, defaultPlugins: [Plugin.Type]) -> [Plugin.Type] {
+    
+    private func getPlugins(_ type: PluginType, defaultPlugins: [Plugin.Type]) -> [Plugin.Type] {
         let filteredPlugins = externalPlugins.filter({ $0.type == type })
         return removeDuplicate(filteredPlugins + defaultPlugins)
     }
-
-    fileprivate func removeDuplicate(_ plugins: [Plugin.Type]) -> [Plugin.Type] {
+    
+    private func removeDuplicate(_ plugins: [Plugin.Type]) -> [Plugin.Type] {
```
towerz: Instead of removing duplicates after merge we could use a different structure to store the plugins since we just need to merge and consider priorities when actually loading the plugin and merging with the externals.

We could use a dictionary with the plugin `name` as a key. This way when a duplicate key is registered, the latter would replace the existing plugin as expected.

Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -1,63 +1,46 @@
...
-    fileprivate func loadExternalMediaControl(_ options: Options) {
-        if let externalMediaControl = options[kMediaControl] as? MediaControl.Type {
-            mediaControl = externalMediaControl
-        }
+            "\n - core: \(corePlugins.map({ $0.name }))")
     }
-
+    
     open func addExternalPlugins(_ externalPlugins: [Plugin.Type]) {
-        self.externalPlugins = externalPlugins
```
towerz: Let's change this to `registerPlugin`, as we documented in the plugin loading diagram. This shouldn't apply only to external plugins, but instead all plugins. Built-in plugins should use the same process as externals.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -64,15 +64,20 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         backgroundColor = UIColor.black
 
         #if os(iOS)
-        mediaControl = loader.mediaControl.create()
+        if let externalMediaControl = options[kMediaControl] as? MediaControl.Type {
```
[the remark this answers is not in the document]
nabilsafatli: Hey @towerz ,

Yeah, that's exactly the idea. But we have another PR to take care of this old media control structure. For now, the goal of this code is to create an interface in the Player that allows the creation of Third Party Plugins using a Singleton Loader.

You can find the other PR about the Media control here: https://github.com/clappr/clappr-ios/pull/227

As soon as this is merged, we will follow what we agreed in the current architecture.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -86,8 +91,8 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         }
     }
 
-    fileprivate func loadPlugins(_ loader: Loader) {
-        for plugin in loader.corePlugins {
+    fileprivate func loadPlugins() {
+        for plugin in Loader.shared.corePlugins {
```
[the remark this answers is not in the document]
nabilsafatli: Hey @towerz , why do we need this inversion of responsibility?

I made the change, please take a look into the implementation of the methods in the `Loader`:
```swift
loadPlugins(in container: Container) {}
//and
loadPlugins(in core: Core) {}
```

I just would like to register the idea here in the comment as well 😄

Sources/Clappr/Classes/Base/Loader.swift
nabilsafatli: Done!

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -155,16 +155,25 @@ open class Player: BaseObject {
...
+    private func unbindPlaybackEvents() {
         for id in playbackEventsListenIds {
             stopListening(id)
         }
 
         playbackEventsListenIds.removeAll()
     }
+    
+    public static func register(plugins: [Plugin.Type]) {
+        if(!hasAlreadyRegisteredPlugins) {
```
leandroalonso: Let's follow the Swift convention: there's a space after the `if` and we don't use brackets.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -155,16 +155,25 @@ open class Player: BaseObject {
...
+            var builtInPlugins: [Plugin.Type] = [AVFoundationPlayback.self]
+            #if os (iOS)
```
leandroalonso: Adding a line before this macro helps to see it. What do you guys think?

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -155,16 +155,25 @@ open class Player: BaseObject {
...
+            builtInPlugins.append(contentsOf: [PosterPlugin.self, SpinnerPlugin.self])
+            #endif
```
leandroalonso: ...and after. Same as above.

paulogama14:
```
Failing tests:
-[WMPCDNRouterRequestTests WMPCDNRouterRequestTests_WhenPerfomingRequestToServer_ShouldReturnAValidCdnNameWhenSuccessfull]
```

Sources/Clappr_iOS/Classes/Base/Player.swift
nabilsafatli: Done!

Sources/Clappr_iOS/Classes/Base/Player.swift
nabilsafatli: Done!

Sources/Clappr_iOS/Classes/Base/Player.swift
nabilsafatli: Done!

Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -33,14 +31,13 @@ open class Container: UIBaseObject {
...
-    public init(loader: Loader = Loader(), options: Options = [:]) {
+    public init(options: Options = [:]) {
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
         self.options = options
-        self.loader = loader
         super.init(frame: CGRect.zero)
         self.sharedData.container = self
         backgroundColor = UIColor.clear
-        loadPlugins()
+        Loader.shared.loadPlugins(in: self)
```
towerz: Just a suggestion for a later change. Perhaps we should isolate the responsibility of instantiating the plugins in another component, like a factory or factory method, and make the component initialization cleaner.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -64,15 +64,20 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
...
+            mediaControl = MediaControl.create()
+        }
         mediaControl?.core = self
+        
         addTapRecognizer()
         #endif
 
         bindEventListeners()
-        loadPlugins(loader)
+        Loader.shared.loadPlugins(in: self)
```
towerz: Like in the `Container` constructor, it would be better to isolate plugin loading calls into a separate component.

Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -1,63 +1,46 @@
...
-    internal(set) open var mediaControl: MediaControl.Type = MediaControl.self
+    public static let shared = Loader()
+    public var plugins: [String: Plugin.Type] = [:]
 
-    fileprivate var externalPlugins = [Plugin.Type]()
-
-    public convenience init() {
-        self.init(externalPlugins: [])
+    var playbacks: [Plugin.Type] {
+        return plugins.filter { $0.value.type == .playback }.map { return $0.value }
```
towerz: There's a slight issue here regarding the playbacks. A `Playback` isn't a regular plugin since only a single one is loaded for a given `Container`. Another behavior that's different is that it's loaded on demand depending on the source we need to play and that there should be a fixed order in which the playbacks are checked.

For these reasons, I belive it's better to have the playbacks actually separated from the actual plugins. So instead of using a computed property here, we should have the playbacks stored in its own stored property, as an array.

Given that change, the `Loader` will need to have a separate `registerPlayback` method, and some other changes. Since it's a big change, we could have it done in a later PR.

nabilsafatli: I totally agree, we will create another PR with these changes.
We have other PRs for the Media Control that need the code developed here, are you guys ok to approve and analyse the next PR with the last comments made here?

PR #227, opened 2018-10-02
Create MediaControl with panels

paulogama14:
### Goal

Create `MediaControl` to conform to architecture 

### How to test

- Run `make test` and make sure that all tests pass.
- Run the example.

### Tech Notes

Now MediaControl is a UICorePlugin.

There are commented tests in `MediaControlTests` that need to uncomment or redo as the plugins are moving.

We are using [media_control](https://github.com/clappr/clappr-ios/tree/feature/media_control) proxy branch. All cards next should be using it

### Out of scope 

In this PR there aren't MediaControl plugins yet, so the media control view is blank.


Sources/Clappr/Classes/Base/Loader.swift
```diff
@@ -18,25 +16,24 @@ open class Loader {
...
 
         Logger.logInfo("plugins:" +
             "\n - playback: \(playbackPlugins.map({ $0.name }))" +
             "\n - container: \(containerPlugins.map({ $0.name }))" +
-            "\n - core: \(corePlugins.map({ $0.name }))" +
-            "\n - mediaControl: \(mediaControl)", scope: "\(type(of: self))")
+            "\n - core: \(corePlugins.map({ $0.name }))", scope: "\(type(of: self))")
     }
 
-    fileprivate func loadExternalMediaControl(_ options: Options) {
```
leandroalonso: I believe that we've removed a Clappr feature: to pass your own `MediaControl`.

If this change goes to production all of the customized `MediaControl` will stop to work. Bringing this feature back would be as simples as return these deleted lines, right?

What do you guys think?

Sources/Clappr/Classes/Enum/Event.swift
```diff
@@ -30,4 +30,5 @@ public enum Event: String {
     case seekableUpdate
     case didChangeDvrAvailability
     case didUpdateOptions
+    case willShowMediaControl
```
leandroalonso: I believe there are a few Events missing here:

* `didShowMediaControl`
* `willHideMediaControl`
* `didHideMediaControl`

Plus, we need to update the [documentation here](https://github.com/clappr/clappr-ios/wiki/Events) with all these new events.

Since this is a branch that will merge into a long living branch I'm not sure about what to do.

Sources/Clappr/Classes/Plugin/Core/MediaControl/Layout/LayoutConstants.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+enum LayoutConstantType {
+    case size, insets
+}
+
+struct LayoutConstants {
```
leandroalonso: Why do we need this?

Sources/Clappr/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -0,0 +1,236 @@
+import Foundation
+
+class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
```
leandroalonso: If we're going to support 3d party `MediaControl` again I highly suggest to make `MediaControl` a more generic class.

Then extend it using this one.

Sources/Clappr/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @objc func tapped() {
+        hideAndStopTimer()
+    }
+
+    override func render() {
+        self.isHidden = true
+        view = UIView()
+        view.backgroundColor = UIColor.clapprBlack60Color()
```
leandroalonso: The Clappr `MediaControl` don't has a translucid black background. Instead, it's a light gradient at the bottom.

You can see here: http://clappr.io/demo/

Plus, we can also check if there's any way to customize this.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -0,0 +1,510 @@
...
+                beforeEach {
+                    coreStub = CoreStub()
+
+                    mediaControl = MediaControl(context: coreStub)
+                    mediaControl.animationDuration = 0.1
+                    mediaControl.secondsToHideControlFast = 0.1
+                    mediaControl.secondsToHideControlSlow = 0.1
+                    mediaControl.render()
+                }
+                /*
```
leandroalonso: We can delete these lines.

It doesn't make sense to leave it here. I don't think we will have any "settings" plugin.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -7,7 +7,6 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         }
     }
     @objc fileprivate(set) open var containers: [Container] = []
-    @objc fileprivate(set) open var mediaControl: MediaControl?
```
leandroalonso: It makes sense to make `mediaControl` not visible in the `Core` anymore?

We're sure that all communication will be made through events?

nabilsafatli:
Hey guys,
Those tests are failing:
```
FullscreenUITests._Fullscreen__when_the_option_of_fullscreen_controled_by_app_is_passed__and_user_taps_on_fullscreen_button__sets_the_player_as_fullscreen()
PlaybackTests._AVFoundationPlayback__when_video_is_playing__is_in_playing_state
```

nabilsafatli:
Hey guys, I guess that the `MediaControlView.swift` was not added to the Xcode project.
screen shot 2018-10-08 at 10 57 09

Sources/Clappr/Classes/Base/Core.swift
paulogama14: Yes, the idea is to remove the couple between Core and MediaControl. With this implementation, we adopt the communication to be made through events.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
paulogama14: sure. 

=]

Sources/Clappr/Classes/Plugin/Core/MediaControl/MediaControl.swift
dezouzart: We know clappr's UI is different then steve, however if we want to keep the same looks as now we'd have to invest much more time. We thought that we agreed to make it simple: make steve's UI and just change assets. We can discuss this with the team with u want to = ) 
But for now I would suggest to move on and create another card to adjust clappr's UI if needed.

PR #228, opened 2018-10-05
Create Fullscreen plugin

paulogama14:
⚠️ DO NOT MERGE THIS PR UNTIL [THIS PR](https://github.com/clappr/clappr-ios/pull/227) IS MERGED ⚠️

### Goal

Creating a Fullscreen plugin to append in MediaControl plugins

### How to test

Run example and tests: 

1. Make sure that Fullscreen functionality conform to business rules
2. Make sure to test Fullscreen controlled by app and by player
3. Make sure to test entering in Fullscreen mode

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -97,8 +97,10 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     }
     
     fileprivate func bindEventListeners() {
+        #if os(iOS)
         listenTo(self, eventName: InternalEvent.userRequestEnterInFullscreen.rawValue) { [weak self] _ in self?.fullscreenHandler?.enterInFullscreen() }
         listenTo(self, eventName: InternalEvent.userRequestExitFullscreen.rawValue) { [weak self] _ in self?.fullscreenHandler?.exitFullscreen() }
+        #endif
     }
     
     fileprivate func renderInContainerView() {
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -62,11 +62,8 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
 
         backgroundColor = UIColor.black
 
-        #if os(iOS)
         addTapRecognizer()
         bindEventListeners()
-        #endif
-        
         loadPlugins(loader)
 
         containers.append(Container(loader: loader, options: options))
@@ -91,9 +88,11 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
     }
 
     fileprivate func addTapRecognizer() {
+        #if os(iOS)
         let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(didTappedView))
         tapRecognizer.delegate = self
         addGestureRecognizer(tapRecognizer)
+        #endif
     }
     
     fileprivate func bindEventListeners() {
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -64,9 +64,9 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
 
         #if os(iOS)
         addTapRecognizer()
-        #endif
-
         bindEventListeners()
+        #endif
+        
         loadPlugins(loader)
 
         containers.append(Container(loader: loader, options: options))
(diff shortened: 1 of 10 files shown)


Sources/Clappr/Classes/Plugin/Core/MediaControl/FullscreenButton.swift
```diff
@@ -0,0 +1,75 @@
+open class FullscreenButton: MediaControlPlugin {
+    private var icon = UIImage.fromName("fullscreen", for: FullscreenButton.self)
```
leandroalonso: The previous fullscreen button toggled between two icons. One displayed when in fullscreen and other when in windowed mode.

Can we reproduce the same behavior here?

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -212,6 +214,22 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
         }
     }
 
+    private func loadDefaultPlugins() {
+        if let disableDefaultPlugins = options?[kDisableDefaultPlugins] as? Bool, disableDefaultPlugins {
```
leandroalonso: If we're passing an Option called `kDisableDefaultPlugins` to Clappr I would assume that will disable every and each plugin.

Instead, this just doesn't load the `MediaControl` plugins. I think we can change this option name to represent what it really does.

Plus, the normal behavior in the architecture is to just override plugins that have the same name. This could be the solution... I'm not sure why we need to disable all plugins.

Sources/Clappr/Classes/Plugin/Core/MediaControl/FullscreenButton.swift
paulogama14: Yes, we can.

=]

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
paulogama14: Done.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -212,6 +214,22 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
...
+            return
+        }
+
+        defaultPlugins.forEach { defaultPlugin in
+            addPlugin(defaultPlugin)
+        }
+    }
+
+    private func addPlugin(_ plugin: MediaControlPlugin.Type) {
+        if !plugins.contains(where: { $0.pluginName == plugin.name}) {
```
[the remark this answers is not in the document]
paulogama14: No, the idea is adding a default plugin when there is not a plugin with the same passing by `kMediaControlPlugins`. 
Example.: If you pass an `XYZ` plugin in `kMediaControlPlugins` options and exists a plugin with the same name in default plugins array, the second will not be added and consequently will not be rendered.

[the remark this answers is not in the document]
leandroalonso: That's the point. The second plugin should be the one to be added, not the first one.

Adding the first one you don't have the feature to override plugins, something that is common in the architecture. You can see how this is handled in the `Loader`: https://github.com/clappr/clappr-ios/blob/dev/Sources/Clappr/Classes/Base/Loader.swift#L52

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -64,9 +64,9 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
 
         #if os(iOS)
         addTapRecognizer()
-        #endif
-
         bindEventListeners()
```
leandroalonso: Here we've added the macro `#if os(iOS)` but is always present inside `bindEventListeners`. We can keep just one to make it simple. :)

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -95,14 +95,14 @@ open class Core: UIBaseObject, UIGestureRecognizerDelegate {
         tapRecognizer.delegate = self
         addGestureRecognizer(tapRecognizer)
     }
-
+    
     fileprivate func bindEventListeners() {
         #if os(iOS)
```
leandroalonso: Stated above.

PR #230, opened 2018-10-11
Create PlayButton Plugin

igoriols:
### Goal

Creating a PlayButton plugin to append in MediaControl plugins.

### How to test

Run example and tests: 

Execute the video and make sure that you can play and pause the video.


Tests/Clappr_Tests/PlayButtonTests.swift
```diff
@@ -0,0 +1,244 @@
...
+                    coreStub.activePlayback?.trigger(Event.playing.rawValue)
+
+                    expect(playButton.view.isHidden).to(beFalse())
+                }
+            }
+        }
+    }
+}
+
+extension UIImage {
```
dezouzart: Don't u think it is better to extract this extension to another file?

dezouzart:
There are a few commented MediaControlTests, should we keep commented instead of delete them? why?

Tests/Clappr_Tests/PlayButtonTests.swift
brunovpl: Done.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -212,6 +214,18 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
         }
     }
 
+    private func loadDefaultPlugins() {
```
nabilsafatli: Heys guys,

According to Clappr architecture, we should not do this here. We had a [similar code during Loaders refactor](https://github.com/clappr/clappr-ios/pull/226#discussion_r223385017).

The loader is responsible for appending Plugins into the context not the context itself.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -76,35 +78,41 @@ class MediaControlTests: QuickSpec {
...
-                    if let seekbar = mediaControl.plugins.first(where: { $0.pluginName == Seekbar.name }) as? Seekbar {
-                        seekbar.view.frame = CGRect(x: 0, y: 0, width: 100, height: 100)
-                    }
-                    
-                    _ = mediaControl.gestureRecognizer(UITapGestureRecognizer(), shouldReceive: UITouch())
-                    
-                    expect(viewController.navigationController?.interactivePopGestureRecognizer?.isEnabled).to(beFalse())
-                }
-            }
+             it("toggles the viewcontroller's swipe to back gesture to true") {
```
nabilsafatli: Why is this commented?

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -245,41 +268,6 @@ class MediaControlTests: QuickSpec {
...
-                        coreStub.activePlayback?.trigger(PlayerInternalEvent.willBeginScrubbing.rawValue)
-                        
-                        expect(mediaControl.isHidden).toEventually(beFalse())
-                        expect(mediaControl.alpha).toEventually(equal(1))
-                        expect(mediaControl.hideControlsTimer?.isValid).toEventually(beFalse())
-                    }
-                }
-                
-                context("when didFinishScrubbing") {
-                    it("hides the media control after some time if the video is playing") {
```
nabilsafatli: Why we deleted this test?

Sources/Clappr/Classes/Base/Options.swift
```diff
@@ -17,6 +17,10 @@ public let kMediaControlAlwaysVisible = "mediaControlAlwaysVisible"
 /// List of MediaControl Plugins
 public let kMediaControlPlugins = "mediaControlPlugins"
```
nabilsafatli: Hey guys, 
I know that this wasn't added in this Pull Request. 😅  But we need to understand that the concept of "Media Control Plugins" don't exist.
We have Core Plugins, not something apart for for media control.

If you need to add a MediaControl plugin and this is not possible for now without this option `kMediaControlPlugins`, let's think about rebasing Media Control branches on the Loader Pull request so we can use the method of `Player.register(plugins: plugins)`. Otherwise we are writing new code without following Clappr's architecture.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
nabilsafatli: Nevermind, I noticed that this is from the old Media Control structure.

Tests/Clappr_Tests/PlayButtonTests.swift
```diff
@@ -0,0 +1,234 @@
...
+import Nimble
+
+@testable import Clappr
+
+class PlayButtonTests: QuickSpec {
+    override func spec() {
+        describe(".PlayButton") {
+
+            describe("#init") {
+                it("is an MediaControlPlugin type") {
```
nabilsafatli: Maybe we can have one instance of PlayButton to do all these 4 tests:
- `init`
- `pluginName `
- `panel `
- `position`

Tests/Clappr_Tests/Classes/Extensions/UIImageExtension.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+extension UIImage {
+    func isEqualTo(image: UIImage) -> Bool {
```
nabilsafatli: Hey, this is not a change requirement, but UIImage conforms to the `Equatable` protocol, so we have already `isEqual` method. What do you think about overriding it and implementing it like this:

```swift 

extension UIImage {
    override open func isEqual(_ object: Any?) -> Bool {
        guard let image = object as? UIImage,
         let lhs: Data = UIImagePNGRepresentation(self),
         let rhs: Data = UIImagePNGRepresentation(image) else {
            return super.isEqual(object)
        }

        return lhs == rhs
    }
}
```

Sources/Clappr/Classes/Base/Options.swift
brunovpl: As accorded we'll treat this issue at another card.

Tests/Clappr_Tests/PlayButtonTests.swift
brunovpl: Done.

Tests/Clappr_Tests/Classes/Extensions/UIImageExtension.swift
brunovpl: Done.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
nabilsafatli: Ok, but I highly suggest to not ever leave commented code behind.
We are very likely to forget it later due to priority changing.

nabilsafatli: I was looking that one of the tests was deleted and the other one commented. Why the different decision?

paulogama14: Sure. We remove this part of the code. Thanks =]

PR #232, opened 2018-10-16
AVFoundationPlayback unification

leandroalonso:
# Goal

To merge the `AVFoundationPlayback` of `iOS` and `tvOS` in a single file.

# How to test

1. Run Clappr (tvOS and iOS)
2. Check the events output. It should start with `ready`, and then some `stalled` and `played` should be displayed

# Additional info

We've fixed 2 issues along with this PR:

1. `Paused` was being triggered after a `didComplete` event. This doesn't happen anymore;
2. In Clappr for `tvOS` the app was not able to listen to all events because the play is automatically triggered. Because of that, the app missed the early events (like `ready`, for example). We fixed that by triggering a play in another thread.


Sources/Clappr/Classes/Base/Playback.swift
```diff
@@ -96,7 +96,9 @@ open class Playback: UIBaseObject, Plugin {
         }
 
         #if os(tvOS)
-        play()
+        DispatchQueue.main.asyncAfter(deadline: .now()) {
```
nabilsafatli: Hey,
We are not using `[weak self]` here, maybe this cause a memory leak problem if someone tries to play and dismiss the player right away.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -233,22 +242,47 @@ open class AVFoundationPlayback: Playback {
     fileprivate func setupPlayer() {
         if let asset = self.asset {
             let item: AVPlayerItem = AVPlayerItem(asset: asset)
+            
+            #if os(iOS)
             player = AVPlayer(playerItem: item)
```
nabilsafatli: Guys, I don't think that we need this `kLoop` anymore.
I suggest use the same structure for tvOS and iOS, `AVQueuePlayer` or `AVPlayer`.
Let's go with `AVPlayer` for both? What do you think?

dezouzart:
@leandroalonso , sometimes at the video's end the playback enter in a "infinity loading" state. Try to seek to the video's end right after video starts.

leandroalonso:
@dezouzart As we've discussed this issue is not caused by our modifications and is currently happening in`dev`.

I believe this is happening because we use a `.mp4` in the example, I wasn't able to reproduce the same issue with `HLS`.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
leandroalonso: Removed.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -20,6 +20,15 @@ open class AVFoundationPlayback: Playback {
     private(set) var seekToTimeWhenReadyToPlay: TimeInterval?
 
     @objc internal dynamic var player: AVPlayer?
+    
+    #if os(tvOS)
+    lazy var nowPlayingService: AVFoundationNowPlayingService = {
+        return AVFoundationNowPlayingService()
+    }()
+    
+    fileprivate var playerLooper: AVPlayerLooper?
```
nabilsafatli: This was forgotten here.

PR #233, opened 2018-10-26
Handle MediaControl plugins in Loader

leandroalonso:
# Goal

To load `MediaControl` plugins through the `Loader`.

# How to test

1. Open the example
2. Play the video
3. You should see fullscreen, time indicator and play/pause button plugins
4. Interact with them and make sure that their behavior is ok


paulogama14:
Hi @leandroalonso, the target Clappr_tvOS doesn't compile. Could you check, please?

leandroalonso:
@paulogama14 Done!

Thanks. :)

PR #236, opened 2018-11-01
Fix tvOS playback

igoriols:
### Goal
Fix tvOS playback.

### How to test
Run tvOS example and check if video is played properly.


nabilsafatli:
Can we cover this scenario with a test so it doesn't happy anymore?

PR #238, opened 2018-11-05
Add tests for the state of the Playback state machine

nabilsafatli:
## Goal

According to the following picture from [Clappr-docs](https://github.com/clappr/clappr-docs/wiki/Architecture), the player should follow some states when interacting with the playback:

With this PR we want to cover some scenarios of the state, our tests are:

#### idle
- when player is ready to play ->  `currentState` must be `idle`
- when play is called -> changes `isPlaying` to *true*
- when pause is called -> changes `currentState` to paused, and `isPaused` to *true*

#### playing
- when pause is called -> it changes `currentState` to paused, and `isPaused` to *true*
- when seek is called -> it changes `currentState` to playing, and `isPlaying` to *true*
- when stop is called -> it changes `currentState` to idle 
- when video is over -> it changes `currentState` to idle
- when is not likely to keep up -> it changes `currentState` to buffering, and `isBuffering` to *true*

##### paused
-  when play is called -> it changes `currentState` to playing, and `isPlaying` to *true*
-  when seek is called -> it keeps `currentState` paused, and `isPaused` *true*
-  when stop is called -> it changes `currentState` to idle
- when is not likely to keep up

#### stalled
- when seek is called -> it keeps `currentState` buffering, and `isBuffering` *true*
- when paused is called ->  it changes `currentState` to paused, and `isPaused` to *true*
- when stop is called -> it changes `currentState` to idle

The next step is to cover the events triggered during the interactions with tests.

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -29,7 +29,7 @@ open class AVFoundationPlayback: Playback {
     
     fileprivate var playerLayer: AVPlayerLayer?
     fileprivate var playerStatus: AVPlayerItemStatus = .unknown
-    fileprivate var currentState = PlaybackState.idle {
+    internal var currentState = PlaybackState.idle {
         didSet {
             switch currentState {
             case .buffering:
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -29,7 +29,7 @@ open class AVFoundationPlayback: Playback {
     
     fileprivate var playerLayer: AVPlayerLayer?
     fileprivate var playerStatus: AVPlayerItemStatus = .unknown
-    internal var currentState = PlaybackState.idle {
+    var currentState = PlaybackState.idle {
         didSet {
             switch currentState {
             case .buffering:
--- a/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
+++ b/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
@@ -945,11 +945,13 @@ class AVFoundationPlaybackTests: QuickSpec {
 
                     context("when play is called") {
                         it("changes isPlaying to true") {
-                            let playback = AVFoundationPlayback()
-                            playback.player = AVPlayerStub()
+                            let options = [kSourceUrl: "http://localhost:8080/sample.m3u8"]
+                            let playback = AVFoundationPlayback(options: options)
+
                             playback.play()
 
                             expect(playback.isPlaying).to(beTrue())
+                            expect(playback.currentState).toEventually(equal(.playing), timeout: 3)
                         }
                     }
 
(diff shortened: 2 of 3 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -29,7 +29,7 @@ open class AVFoundationPlayback: Playback {
     
     fileprivate var playerLayer: AVPlayerLayer?
     fileprivate var playerStatus: AVPlayerItemStatus = .unknown
-    fileprivate var currentState = PlaybackState.idle {
+    internal var currentState = PlaybackState.idle {
```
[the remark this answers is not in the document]
nabilsafatli: Done!

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -933,6 +933,207 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                    }
+
+                    context("when play is called") {
+                        it("changes isPlaying to true") {
+                            let options = [kSourceUrl: "http://localhost:8080/sample.m3u8"]
+                            let playback = AVFoundationPlayback(options: options)
+
+                            playback.play()
+
+                            expect(playback.isPlaying).to(beTrue())
```
leandroalonso: Shouldn't we also test what is the value of `isPaused` and `isBuffering` for each case? In that way, we can ensure that every representation of the state is ok.

What do you guys think?

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -933,6 +933,207 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                        }
+                    }
+
+                    context("when pause is called") {
+                        it("changes current state to paused") {
+                            let playback = AVFoundationPlayback()
+
+                            playback.pause()
+                            
+                            expect(playback.isPaused).to(beTrue())
```
leandroalonso: Same as above. `isPlaying` and `isBuffering` must be `false`, right?

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -933,6 +933,207 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+
+                            playback.play()
+                            playback.stop()
+                            
+                            expect(playback.currentState).to(equal(.idle))
+                        }
+                    }
+                    
+                    context("when is likely to keep up") {
+                        it("has isPlaying as true") {
```
leandroalonso: What is the difference between this test and the second `it`? It seems that is the same.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
nabilsafatli: Hey @leandroalonso ,
You are right, we should test those values to make sure about the state representation.
But when we did that, we found that in the moment of the test those properties may have a delay to change their value.

What we did this time, we added a timeout for the expectation to make it possible to test.
Take a look into the last commit and let me know if it's how you expect.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
nabilsafatli: Done.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
nabilsafatli: You are right, we've just removed it.

PR #242, opened 2018-11-14
Events renaming

dezouzart:
### Goal 
To deprecate event names and rename/create new

### How to test
1 - Listen to new events and check if they were called
2 - Run ios and tvos tests

#### bonus
fix load bug which was not updating source option

#### ps
wiki will be update on release process


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -443,6 +447,7 @@ open class AVFoundationPlayback: Playback {
         }
 
         self.trigger(.airPlayStatusUpdate, userInfo: ["externalPlaybackActive": player!.isExternalPlaybackActive])
+        self.trigger(.didUpdateAirPlayStatus, userInfo: ["externalPlaybackActive": player!.isExternalPlaybackActive])
```
paulogama14: Hey @dezouzart, maybe we can avoid the force unwrapping of `player` object. One option is to use `concretePlayer` object. What do you think?

Sources/Clappr/Classes/Enum/Event.swift
```diff
@@ -34,4 +33,26 @@ public enum Event: String, CaseIterable {
     case didShowMediaControl
     case willHideMediaControl
     case didHideMediaControl
+    @available(*, deprecated, message: "Update to willSeek")
+    case seek
```
igoriols: Hey @dezouzart , seek was already deprecated before as you can see [here](https://github.com/clappr/clappr-ios/wiki/Events) . Maybe we can remove it.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -76,14 +76,22 @@ open class Player: BaseObject {
...
-             Event.seek.rawValue, Event.didSeek.rawValue,
-             Event.subtitleSelected.rawValue, Event.audioSelected.rawValue])
+             Event.airPlayStatusUpdate.rawValue, Event.willSeek.rawValue, Event.didUpdateAirPlayStatus.rawValue,
+             Event.didSeek.rawValue,
+             Event.subtitleSelected.rawValue, Event.audioSelected.rawValue,
+             Event.didFindSubtitle.rawValue, Event.didFindAudio.rawValue,
+             Event.didSelectSubtitle.rawValue, Event.didSelectAudio.rawValue,
+             Event.willDestroy.rawValue, Event.didDestroy.rawValue, Event.willLoadSource.rawValue,
+             Event.didLoadSource.rawValue, Event.willChangePlayback.rawValue, Event.didChangePlayback.rawValue,
+             Event.willChangeActiveContainer.rawValue, Event.didChangeActiveContainer.rawValue,
```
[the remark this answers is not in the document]
igoriols: Fixed! 👍

PR #243, opened 2018-11-14
Docs: Media Control

paulogama14:
### Goal 

Improve docs with the implementation of `MediaControl` by panels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,9 +51,9 @@ github "clappr/clappr-ios"  ~> 0.9.0
 
 Run `carthage update` to build the framework and drag the built `Clappr.framework` into your Xcode project.
 
-## Usage
+# Usage
 
-### iOS
+## iOS
 
 #### Create
 ```swift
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,6 +67,16 @@ let player = Player(options: options)
 player.attachTo(yourView, controller: self)
 ```
 
+#### Media Control
+
+It controls all the user interaction between UI and the playback, it relies only on playback abstraction, therefore, this component should work in playback.
+
+It's a set of `Stack Views` divided into panels (`Top`, `Center` and `Bottom`) and can be used to organize the components on the screen.
```
leandroalonso: I believe we can mention how the `MediaControl` works instead of talking about its implementation. What do you think?

Some key points that we can mention:

1) It's an element that contains plugins which control the playback;
2) You can customize it with your own plugins (aka `Media Control Plugins`)
3) It is divided in panels (like you said)

README.md
```diff
@@ -67,6 +67,16 @@ let player = Player(options: options)
...
+
+#### Media Control Plugins
+
+Allow providing controls for the `Player`. There are default plugins available with basic user interactions like: `Play/Pause`, `TimeIndicator`, `Seekbar` and `Fullscreen`.
```
leandroalonso: The `MediaControl` plugins can be used to control the playback but can be used to other approaches that doesn't involve that: displaying a video's title, for example.

What do you think about providing a better description? Like:

1) It should inherits from `MediaControlPlugin`
2) Should contain the `panel` and `position`
3) Should have a unique name

And the link to the examples (`PlayButton.swift`, `Seekbar.swift` and so on)

README.md
paulogama14: Done!

README.md
paulogama14: Done.

PR #246, opened 2018-11-26
Create Core and Container Factories

nabilsafatli:
# Goal

Refactor to improve *Clappr's Architecture* scalability to test new Playbacks. Where we:

* Changed the way how we instantiate a `Core` and `Container` in `Player`.
* Changed the playbacks that exist in Loader from `[Plugin.Type]` to `[Playback.Type]`

# How to test

- Run all tests
- Check if the video loads properly
- Use `configure` to reload a video
- Use `load` to reload a video

### DIFF
--- a/Sources/Clappr/Classes/Extension/Array+appendOrReplace.swift
+++ b/Sources/Clappr/Classes/Extension/Array+appendOrReplace.swift
@@ -16,3 +16,20 @@ extension Array where Element == Plugin.Type {
         self.append(contentsOf: result)
     }
 }
+
+extension Array where Element == Playback.Type {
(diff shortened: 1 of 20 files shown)


Sources/Clappr/Classes/Extension/Array+appendOrReplace.swift
```diff
@@ -16,3 +16,20 @@ extension Array where Element == Plugin.Type {
         self.append(contentsOf: result)
     }
 }
+
+extension Array where Element == Playback.Type {
```
paulogama14: Maybe we could increment `ArrayAppendOrReplaceTests` with this new extension. What do you think?

PR #247, opened 2018-11-27
Fix ready event

geisonfgf:
### Goal
Adjust the moment when ready event is triggered to be after AVFoundationPlayback is rendered. 

### How to test
1 - Comment the line 597 at file AVFoundationPlayback.swift (with this the video will **not** play).
2 - Run Clappr example app;
3 - Check if ready event appears at log.


igoriols:
Hey @geisonfgf  and @paulogama14, I couldn't play the video. It keeps loading indefinitely.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -287,14 +245,46 @@ open class AVFoundationPlayback: Playback {
...
             object: player?.currentItem)
     }
 
+    @objc public func setDelegate(_ delegate: AVAssetResourceLoaderDelegate) {
+        self.asset?.resourceLoader.setDelegate(delegate, queue: DispatchQueue(label: "\(String(describing: asset?.url))-delegateQueue"))
+    }
+
+    public required init?(coder _: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
```
igoriols: We don't need this constructor anymore.

paulogama14: Okay!

paulogama14:
Hey @igoriols, you are right. When we comment that line the video won't play and this is the expected behavior because it's necessary to test if the ready is triggered when Playback is really ready. Different from the old implementation the ready was triggered on a `play` action.

Example/Clappr/ViewController.swift
```diff
@@ -30,7 +30,10 @@ class ViewController: UIViewController {
 
         player.on(Event.didComplete) { _ in print("on Complete") }
 
-        player.on(Event.ready) { _ in print("on Ready") }
+        player.on(Event.ready) { [weak self] _ in
+            print("on Ready")
+            self?.player.play()
```
leandroalonso: I removed this play here and the player gets stuck in a weird state, showing the spinner.

I believe this is due to errors with the `PosterPlugin` after those changes.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -308,6 +294,26 @@ open class AVFoundationPlayback: Playback {
         }
     }
 
+    open override func render() {
+        if self.startAt != 0.0 && self.playbackType == .vod {
+            self.seek(self.startAt)
```
leandroalonso: The seek should be made right before the video starts playing for the first time. Are we sure this gonna work calling this method in the render?

Tests/Clappr_Tests/AVFoundationPlaybackStateMachineEventsTests.swift
```diff
@@ -29,7 +29,7 @@ class AVFoundationPlaybackStateMachineEventsTests: QuickSpec {
                         let options = [kSourceUrl: "http://localhost:8080/sample.m3u8"]
                         let playback = AVFoundationPlayback(options: options)
                         let expectedEvents: [Event] = [
-                            .ready, .willPlay, .stalling, .willPlay, .playing,
```
leandroalonso: This test is failing in tvOS.

Tests/Clappr_Tests/AVFoundationPlaybackStateMachineEventsTests.swift
```diff
@@ -62,16 +63,17 @@ class AVFoundationPlaybackStateMachineEventsTests: QuickSpec {
...
```
leandroalonso: Same as above.

Tests/Clappr_Tests/AVFoundationPlaybackStateMachineEventsTests.swift
paulogama14: Done.

Tests/Clappr_Tests/AVFoundationPlaybackStateMachineEventsTests.swift
paulogama14: Done!

Example/Clappr/ViewController.swift
paulogama14: We remove the play from `handleBufferingEvent` method, so we need to call a play explicitly when `ready` has listened. What do you think @leandroalonso?

leandroalonso: No problem with this change. :)

What I do believe we need to do is to keep the player consistent, doesn't matter which use case. If the app doesn't play the video right on `ready` an infinite spinner will appear. I believe we can fix this by making the `MediaControl` prepared for this case.

geisonfgf:
We changed the approach of this solution so it was created a new PR with it https://github.com/clappr/clappr-ios/pull/251

PR #256, opened 2018-12-10
Fix warnings

geisonfgf:
### Goal
Remove project warnings

### How to Steve
Build and check if warnings are removed in all targets


paulogama14:
We create another [PR](https://github.com/clappr/clappr-ios/pull/257) to cover this scenario.

PR #261, opened 2019-01-02
Return bitrate information from Playback

geisonfgf:
# Goal
Return bitrate information from Playback.

# How to test
In the example app past the code below in the `ViewController.swift` at line 24
```swift
player.on(Event.playing) { _ in
    print("on Play")
    if let playback = self.player.activePlayback as? AVFoundationPlayback {
        print("Bitrate: \(String(describing: playback.getBitrate()))")
        print("Avg Bitrate: \(String(describing: playback.getAvgBitrate()))")
    }
}
```
And check in the log if the values are returned.

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -697,3 +697,23 @@ extension AVFoundationPlayback {
         return 5
     }
 }
+
+// MARK: - Metrics
+extension AVFoundationPlayback {
+    public func getBitrate() -> Double? {
+        guard let logEvent = player?.currentItem?.accessLog()?.events.last as? AVPlayerItemAccessLogEvent else { return nil }
+        if (logEvent.segmentsDownloadedDuration ?? 0.0) > 0 {
+            return logEvent.indicatedBitrate
+        }
+        return logEvent.observedBitrate
+    }
+
+    public func getAvgBitrate() -> Double? {
+        guard let logEvent = player?.currentItem?.accessLog()?.events.last as? AVPlayerItemAccessLogEvent else { return nil }
+        if #available(iOS 10.0, *) {
+            return logEvent.averageVideoBitrate
+        } else {
+            return nil
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -697,3 +697,22 @@ extension AVFoundationPlayback {
         return 5
     }
 }
+
+// MARK: - Metrics
+extension AVFoundationPlayback {
+    public func getBitrate() -> Double? {
+        guard let logEvent = player?.currentItem?.accessLog()?.events.last else { return nil }
```
caiobzen: how about having a function that returns the last `AVPlayerItemAccessLogEvent`? So we can reuse it in both `getBitrate` and `getAvgBitrate` functions. Something like this:
```swift
private func lastLogEvent() -> AVPlayerItemAccessLogEvent? {
    return player?.currentItem?.accessLog()?.events.last
}

public func getBitrate() -> Double? {
    guard let logEvent = lastLogEvent() else { return nil }
    if logEvent.segmentsDownloadedDuration > 0 {
        return logEvent.indicatedBitrate
    }
    return logEvent.observedBitrate
}

public func getAvgBitrate() -> Double? {
    guard let logEvent = lastLogEvent else { return nil }
    if #available(iOS 10.0, *) {
        return logEvent.averageVideoBitrate
    }
    return nil
}
```

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -697,3 +697,26 @@ extension AVFoundationPlayback {
         return 5
     }
 }
+
+// MARK: - Metrics
+extension AVFoundationPlayback {
```
leandroalonso: Can we move this part to a file named `AVFoundationPlayback+Bitrate.swift`? That way we can avoid this "MARK" comment.

geisonfgf: Done ;-)

PR #262, opened 2019-01-07
Ordering media control plugins

geisonfgf:
# Goal
Order Media Control plugins based in the kMediaControlPluginsOrder option.

# How to test
In the Clappr example app `ViewController.swift` add the option before instantiate Player: `options[kMediaControlPluginsOrder] = "TimeIndicator,FullscreenButton"` 
Change the TimeIndicatorPlugin and FullscreenButton to same position, for example left, run the app and check if the order are correct.

**P.S.** Remind that the order of disposing the plugins in media control follow the position so if position is left the first plugin mentioned in the option came to left and if the position was right the first plugin will be displayed at the right.

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -120,16 +120,25 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
 
     #if os(iOS)
     private func renderCoreAndMediaControlPlugins() {
-        plugins.forEach { plugin in
-            if isNotMediaControlPlugin(plugin) {
-                view.addSubview(plugin.view)
-                plugin.render()
-            }
+        renderCorePlugins()
+        renderMediaControlPlugins()
+    }
 
+    private func renderCorePlugins() {
+        plugins.filter { isNotMediaControlPlugin($0) }.forEach { plugin in
+            view.addSubview(plugin.view)
+            plugin.render()
+        }
+    }
 
-            if plugin is MediaControl, let mediaControl = plugin as? MediaControl {
-                mediaControl.renderPlugins(plugins)
-            }
+    private func renderMediaControlPlugins() {
+        let mediaControl = plugins.first { $0 is MediaControl }
+
+        if let mediaControl = mediaControl as? MediaControl {
+            let mediaControlPlugins = plugins
+                .filter { $0 is MediaControlPlugin }
+                .map { $0 as! MediaControlPlugin }
+            mediaControl.renderPlugins(mediaControlPlugins)
         }
     }
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -135,9 +135,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         let mediaControl = plugins.first { $0 is MediaControl }
 
         if let mediaControl = mediaControl as? MediaControl {
-            let mediaControlPlugins = plugins
-                .filter { $0 is MediaControlPlugin }
-                .map { $0 as! MediaControlPlugin }
+            let mediaControlPlugins = plugins.compactMap { $0 as? MediaControlPlugin }
             mediaControl.renderPlugins(mediaControlPlugins)
         }
     }
--- a/Sources/Clappr/Classes/Base/Options.swift
+++ b/Sources/Clappr/Classes/Base/Options.swift
@@ -16,7 +16,7 @@ public let kMediaControlAlwaysVisible = "mediaControlAlwaysVisible"
 
 // List of MediaControl Plugins
 public let kMediaControlPlugins = "mediaControlPlugins"
-// Order of MediaControlPlugins
+// Order of MediaControl Plugins
 public let kMediaControlPluginsOrder = "mediaControlPluginsOrder"
 
 public let kLoop = "loop"
--- a/Sources/Clappr/Classes/Base/Options.swift
+++ b/Sources/Clappr/Classes/Base/Options.swift
@@ -16,6 +16,8 @@ public let kMediaControlAlwaysVisible = "mediaControlAlwaysVisible"
 
 // List of MediaControl Plugins
 public let kMediaControlPlugins = "mediaControlPlugins"
+// Order of MediaControlPlugins
+public let kMediaControlPluginsOrder = "mediaControlPluginsOrder"
 
 public let kLoop = "loop"
 public let kMetaData = "metadata"
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
@@ -219,10 +219,23 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
     }
 
     func renderPlugins(_ plugins: [MediaControlPlugin]) {
-        plugins
-            .forEach { plugin in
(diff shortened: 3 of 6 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -218,13 +218,24 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
...
+        if let mediaControlPluginOrderOption = core?.options[kMediaControlPluginsOrder] as? String {
+            let pluginsOrder = mediaControlPluginOrderOption
+                .replacingOccurrences(of: " ", with: "")
+                .components(separatedBy: ",")
+
+            var orderedPlugins: Array<MediaControlPlugin> = plugins.filter { pluginsOrder.contains($0.pluginName)}
+            orderedPlugins.append(contentsOf: plugins.filter { !pluginsOrder.contains($0.pluginName)})
+
+            orderedPlugins.forEach { plugin in
+                mediaControlView.addSubview(plugin.view, panel: plugin.panel, position: plugin.position)
```
caiobzen: Maybe for later since other places use the exact same function call, but how about having another `addSubview` or even `addPlugin` function that takes a plugin instead of taking its properties? so we end up calling it like `mediaControlView.addPlugin(plugin)`

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -218,13 +218,24 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
...
+                plugin.render()
+            }
+        } else {
+            plugins.forEach { plugin in
```
caiobzen: In my opinion, we could try to come up with a solution that ends up with a single:

```swift
plugins.forEach { plugin in
    mediaControlView.addSubview(plugin.view, panel: plugin.panel, position: plugin.position)
    plugin.render()
}
```

We could have this first step of checking the plugins order property, do all the sorting operation and then call the addSubview with whichever array of plugins we have.

My two cents: have another function called `sortPluginsIfNeeded` that returns the plugins in the right order, and then add them up to the interface:

```swift
let sortedPlugins = sortPluginsIfNeeded(plugins)
sortedPlugins.forEach { plugin in
    mediaControlView.addSubview(plugin.view, panel: plugin.panel, position: plugin.position)
    plugin.render()
}
```

or even pass the already sorted plugins to this `renderPlugins` function. 
Too many options, sorry! :joy:

leandroalonso: ...or we can move to a Array of `MediaControlPlugin` extension. :)

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -120,16 +120,25 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
+            view.addSubview(plugin.view)
+            plugin.render()
+        }
+    }
 
-            if plugin is MediaControl, let mediaControl = plugin as? MediaControl {
-                mediaControl.renderPlugins(plugins)
-            }
+    private func renderMediaControlPlugins() {
+        let mediaControl = plugins.first { $0 is MediaControl }
```
paulogama14: What is the behavior when we pass more than one Media Control? I mean, this architecture allows us to pass a Media Control in the plugins array and I'm not sure which one will be called.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -120,16 +120,25 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
+
+        if let mediaControl = mediaControl as? MediaControl {
+            let mediaControlPlugins = plugins
+                .filter { $0 is MediaControlPlugin }
+                .map { $0 as! MediaControlPlugin }
+            mediaControl.renderPlugins(mediaControlPlugins)
```
paulogama14: Why are we go through in two distinct moments in plugins array? Can we keep one like filter or map?

Sources/Clappr/Classes/Base/Options.swift
```diff
@@ -16,6 +16,8 @@ public let kMediaControlAlwaysVisible = "mediaControlAlwaysVisible"
 
 // List of MediaControl Plugins
 public let kMediaControlPlugins = "mediaControlPlugins"
+// Order of MediaControlPlugins
```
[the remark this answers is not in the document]
geisonfgf: Done

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -472,19 +472,20 @@ class MediaControlTests: QuickSpec {
...
                     }
                 }
-                
-                context("when passing a plugin that is not MediaControlPlugin") {
-                    it("does not add it to the view") {
-                        let plugins = [MediaControlPluginMock(), UICorePlugin()]
+
+                context("when kMediaControlPluginsOrder is passed") {
+                    it("renders the plugins following the kMediaControlPluginsOrder order") {
+                        core.options[kMediaControlPluginsOrder] = "FullscreenButton"
```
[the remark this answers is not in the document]
geisonfgf: Done

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
geisonfgf: We changed the parameters names only to have a more meaningful swifty names but add the method addPlugin to the view will give to view a responsibility out of view context so we preferred do not change it at this moment. Do you agree?

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
geisonfgf: Done

Sources/Clappr/Classes/Base/Core.swift
geisonfgf: Done

Sources/Clappr/Classes/Base/Core.swift
geisonfgf: Well, at this moment it was assumed that only one MediaControl will be passed but if it changes and become necessary to handle multiple MediaControls it will be needed a more big and specific refactor to fix that.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
caiobzen: Totally! Good point :+1:

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -472,19 +472,21 @@ class MediaControlTests: QuickSpec {
...
-                        mediaControl.mediaControlView = mediaControlViewMock
                         mediaControl.render()
 
-                        
                         mediaControl.renderPlugins(plugins)
-                        
-                        expect(mediaControlViewMock.didCallAddSubviewWithView).to(equal(plugins.first?.view))
 
+                        let bottomRightView = mediaControl.mediaControlView.bottomRight
+                        let pluginView = bottomRightView?.subviews[0]
```
caiobzen: `subviews.first` instead of `subviews[0]` maybe?

geisonfgf: Done

PR #265, opened 2019-01-10
Change the way plugin order option is passed

geisonfgf:
# Goal
Order Media Control plugins based in the kMediaControlPluginsOrder option.

# How to test
In the Clappr example app `ViewController.swift` add the option before instantiate Player: `options[kMediaControlPluginsOrder] = ["TimeIndicator", "FullscreenButton"]` 
Change the TimeIndicatorPlugin and FullscreenButton to same position, for example left, run the app and check if the order are correct.

**P.S.** Remind that the order of disposing the plugins in media control follow the position so if position is left the first plugin mentioned in the option came to left and if the position was right the first plugin will be displayed at the right.

### DIFF
--- a/Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
+++ b/Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
@@ -474,19 +474,34 @@ class MediaControlTests: QuickSpec {
                 }
 
                 context("when kMediaControlPluginsOrder is passed") {
-                    it("renders the plugins following the kMediaControlPluginsOrder order") {
-                        let pluginName = "FullscreenButton"
-                        core.options[kMediaControlPluginsOrder] = [pluginName]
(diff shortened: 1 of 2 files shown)


Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/MediaControlTests.swift
```diff
@@ -476,7 +476,7 @@ class MediaControlTests: QuickSpec {
                 context("when kMediaControlPluginsOrder is passed") {
                     it("renders the plugins following the kMediaControlPluginsOrder order") {
                         let pluginName = "FullscreenButton"
-                        core.options[kMediaControlPluginsOrder] = pluginName
+                        core.options[kMediaControlPluginsOrder] = [pluginName]
```
jademcosta: Do you think we should add more tests, for cases where we have more than 1 plugin name being passed as option? Or do you think this one is enough?

igoriols: Hey @jademcosta, we just did it.

PR #267, opened 2019-01-17
Change core visibility in player

BruunoMM:
Goal
===
Change the visibility of the Player's core.

How to test
===
1. Run iOS and tvOS unit tests.
2. Run the sample app and check if everything works as expected.


Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -140,6 +150,14 @@ open class Player: BaseObject {
     @objc open func setFullscreen(_ fullscreen: Bool) {
         core?.setFullscreen(fullscreen)
     }
+    
+    open var options: Options? {
+        return core?.options
+    }
+    
+    open var corePlugins: [UICorePlugin]? {
```
jademcosta: After talking with some people, I realised we don't want to expose the plugin list. Instead, we could create a method that returns the instance of the plugin if the name is provided. Web player has it implemented here: https://github.com/clappr/clappr/blob/dev/src/components/player.js#L576

BruunoMM: It's done.
We implemented the method `getPlugin`, if you wanna take a look... 😜

Tests/Clappr_Tests/PlayerTests.swift
```diff
@@ -236,7 +236,7 @@ class PlayerTests: QuickSpec {
                         Player.register(plugins: [LoggerPlugin.self])
                         player = Player(options: options)
 
-                        let loggerPlugin = player.core?.plugins.first { $0 is LoggerPlugin }
+                        let loggerPlugin = player.getPlugin(name: LoggerPlugin.name)
```
jademcosta: I was thinking about adding a test to check if it can return the correct plugin if the type of the plugin is `container`, but I think I'm being too picky. I'll approve it and let you decide if you create this test or not ;)

PR #268, opened 2019-01-18
Prevent Player crash when Plugin throw exception

leandroalonso:
# Goal

Prevent Player crash when Plugin throw exception during `render`, `destroy` or event callbacks.

# How to test

1. In `ViewController.swift` add this plugin:

```swift
class BrokenPlugin: UICorePlugin {
    override var pluginName: String {
        return "broken"
    }

    required public init(context: UIObject) {
        super.init(context: context)
        bindEvents()
    }

    required public init() {
        super.init()
    }

    private func bindEvents() {
        stopListening()

        bindCoreEvents()
    }

    open func bindCoreEvents() {
        if let core = core {
            listenTo(core, eventName: Event.didChangeActiveContainer.rawValue) { [weak self] _ in self?.iWillCrash() }
        }
    }

    override func destroy() {
        iWillCrash()
    }

    override func render() {
        iWillCrash()
    }

    func iWillCrash() {
        NSException(name:NSExceptionName(rawValue: "TestError"), reason:"Test Error", userInfo:nil).raise()
    }
}
```

2. And before the line `17` (the player instantiation) add this:

```swift
Player.register(plugins: [BrokenPlugin.self])
```

3. Run the player. It should not crash and output some errors in console.

4. Change `UICorePlugin` to `MediaControlPlugin` then do the same test
5. Change to `UIContainerPlugin` and test

# Wait! What about Swift exceptions?

Our methods (`render`, `destroy` and event callbacks) does not allow the throw of Swift exceptions. If the user wants to use them he will have to threat'em inside the scope.

Also, `try!` and `fatalError` triggers fatal and non-catchable exceptions. In this case, the app will crash and there's nothing we can do.


Tests/Clappr_Tests/BaseObjectTests.swift
```diff
@@ -70,6 +70,19 @@ class BaseObjectTests: QuickSpec {
 
                     expect(callbackWasCalled) == false
                 }
+
+                it("protect the main thread when plugin crashes in render") {
+                    let expectation = QuickSpec.current.expectation(description: "doesn't crash")
+                    let baseObject = BaseObject()
+
+                    baseObject.on(eventName) { _ in
```
caiobzen: Should we move the `baseObject.on(eventName) { _ in ... }` one line up with the rest of the `Arrange` phase?
It looks like part of it, please correct me if I'm wrong.

leandroalonso: I will move it, thanks!

leandroalonso:
@igoriols makes sense. I will change it, thanks!

PR #276, opened 2019-02-05
Double tap plugin

caiobzen:
## 🏁 Goal

- To add the doubleTapPlugin to the Player
    - It should work while displaying or not the mediaControl

## ✅ How to test

Besides running all the tests, please do:

1. Run the app
2. Double tap on the right side of the view
3. Double tap on the left side of the view
4. Tap to show the media controls
5. Repeat step 2 and 3


Sources/Clappr_iOS/Classes/Animation/DoubleTapAnimation.swift
```diff
@@ -0,0 +1,254 @@
...
+    private var rightBubbleView = UIView()
+    
+    private var leftBubbleHeight = NSLayoutConstraint()
+    private var leftBubbleWidth = NSLayoutConstraint()
+    
+    private var rightBubbleHeight = NSLayoutConstraint()
+    private var rightBubbleWidth = NSLayoutConstraint()
+    
+    init(_ core: Core?) {
+        self.core = core
```
leandroalonso: Wouldn't be better if this class just represented the view itself?

A view that contained the left bubble and the right bubble?

The logic of adding the view, bringing to the front and so on would be made anywhere else. Here we coupled this class to the Core and if in the future we want to move it to anywhere else we would have to change it.

Making it not-specialized will give us the flexibility of use it anywhere we want. What do you think?

Sources/Clappr_iOS/Classes/Animation/DoubleTapAnimation.swift
```diff
@@ -0,0 +1,254 @@
...
+        return view
+    }
+    
+    private func setupBubble(_ bubble: UIView, within view: UIView, heightConstraint: inout NSLayoutConstraint, widthConstraint: inout NSLayoutConstraint, position: NSLayoutAttribute) {
+        
+        bubble.backgroundColor = UIColor(white: 0, alpha: 0.2)
+        view.addSubview(bubble)
+        bubble.translatesAutoresizingMaskIntoConstraints = false
+        
+        view.addConstraint(NSLayoutConstraint(item: bubble,
```
leandroalonso: There's always the endless discussion about viewcode or xibs. Maybe in this case we could have reduced the amount of code by using a nib.

We do not need to do that now but it's a discussion that we can have later.

Sources/Clappr_iOS/Classes/Plugin/Base/DoubleTapPlugin.swift
```diff
@@ -15,11 +17,36 @@ public class DoubleTapPlugin: UICorePlugin {
...
+    
+    private func bindCoreEvents() {
+        guard let core = core else { return }
+        listenTo(core, eventName: Event.didChangeActiveContainer.rawValue) { [weak self] _ in self?.bindEvents() }
+        listenTo(core, eventName: Event.didShowModal.rawValue) { [weak self] _ in self?.removeDoubleTapGesture() }
+        listenTo(core, eventName: Event.didHideModal.rawValue) { [weak self] _ in self?.addDoubleTapGesture() }
+    }
+    
+    func removeDoubleTapGesture() {
+        Logger.logInfo(#function)
```
jademcosta: I got confused with this one. Is this supposed to be a `log` or should it remove the behaviour when a modal is open?

Sources/Clappr_iOS/Classes/Animation/DoubleTapAnimation.swift
```diff
@@ -0,0 +1,254 @@
...
+        animate(leftBubbleView, parentView: view, widthConstraint: leftBubbleWidth, heightConstraint: leftBubbleHeight)
+        animate(backLabel)
+        animate(backIcon3, delay: 0)
+        animate(backIcon2, delay: 0.2)
+        animate(backIcon1, delay: 0.4)
+    }
+    
+    func animateForward() {
+        guard let playback = core?.activePlayback,
+            playback.position + 10 < playback.duration,
```
jademcosta: What do you think about moving this check to the `Plugin` class? 
OI have the impression it has more in common with plugin rules than it has to do with `Animation`.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/DoubleTapMediaControlPlugin.swift
```diff
@@ -0,0 +1,49 @@
+import UIKit
+
+public class DoubleTapMediaControlPlugin: DoubleTapPlugin {
```
jademcosta: Is it ok for a plugin that will be added to media control, to be a descendant from `UICorePlugin`?

Sources/Clappr_iOS/Classes/Plugin/Base/DoubleTapPlugin.swift
```diff
@@ -0,0 +1,70 @@
...
```
jademcosta: Is the plugin usable without adding and removing the gesture recogniser?
What I want to clarify with this question is if we should let the log, or if we should throw an error telling this should be implemented by descendant classes.

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/DoubleTapMediaControlPluginTests.swift
```diff
@@ -0,0 +1,84 @@
...
+                    }
+                }
+                
+                context("and it colides with another UICorePlugin") {
+                    it("should not seek") {
+                        mediaControl.renderPlugins([playButton])
+                        playButton.view.layoutIfNeeded()
+                        mediaControl.view.layoutIfNeeded()
+                        
+                        let shouldSeek = doubleTapPlugin.shouldSeek(point: CGPoint(x: 100, y: 100))
```
jademcosta: Is the `shouldSeek` method visible just for this test, or does it used somewhere else?

I'm asking because if it is the case, would it work if instead of checking using this method, we check the playback time as we did in other tests?

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/DoubleTapMediaControlPluginTests.swift
```diff
@@ -0,0 +1,84 @@
...
+                        doubleTapPlugin.doubleTapSeek(xPosition: core.view.frame.width)
+                        
+                        expect(core.playbackMock?.didCallSeek).to(beTrue())
+                        expect(core.playbackMock?.didCallSeekWithValue).to(equal(30))
+                    }
+                }
+                
+                context("and it colides with another UICorePlugin") {
+                    it("should not seek") {
+                        mediaControl.renderPlugins([playButton])
```
jademcosta: When I removed this line on my local setup, the test still pass.

Sources/Clappr_iOS/Classes/Animation/DoubleTapAnimation.swift
caiobzen: Totally! good catch!

Sources/Clappr_iOS/Classes/Plugin/Base/DoubleTapPlugin.swift
caiobzen: That's an excellent question! I wasn't sure about which approach to take.
Throwing an exception seems more reasonable since we want it to crash if its not properly setup. What do you think?

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/DoubleTapMediaControlPluginTests.swift
caiobzen: Kinda. It's more because we can't simulate a double tap event in order to test it.
We're using the same approach to test `doubleTapSeek` function.
What we can do is: create a big plugin that covers the entire screen, and then ask the plugin to seek. Sounds good?

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/DoubleTapMediaControlPlugin.swift
caiobzen: It seems to be an architectural limitation, @jademcosta. Besides having MediaControl on its name, this plugin is added to core instead.

Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
```diff
@@ -0,0 +1,224 @@
...
+    var bubbleWidth = NSLayoutConstraint()
+    var bubbleSide: SeekBubbleSide!
+    
+    func setup(within parentView: UIView, bubbleSide: SeekBubbleSide, text: String = "10 segundos", numberOfIndicators: Int = 3) {
+        self.parentView = parentView
+        self.bubbleSide = bubbleSide
+        self.text = text
+        parentView.clipsToBounds = true
+        
+        setupBubble(within: parentView, heightConstraint: &bubbleHeight, widthConstraint: &bubbleWidth, position: bubbleSide.position())
```
leandroalonso: If the `bubbleHeight` and `bubbleWidth` are properties of this class, why to not just access it using `self`?

Tests/Clappr_Tests/Classes/Plugins/Core/MediaControl/DoubleTapMediaControlPluginTests.swift
jademcosta: Makes sense :)

PR #277, opened 2019-02-08
Remove Kingfisher

paulogama14:
# Goal
 
Remove Kingfisher dependency from the project.

# Additional info

- `PosterPlugin` receives an URL to show an image when the video was not started or has finished. 
- With actual implementation, we are making the same request three times (didUpdateOptions / requestPosterUpdate events and renderPlugin) and we will not fix in this PR:

# How to Test

**First way**
Run `Clappr_Example` and check if [this image](http://clappr.io/poster.png) was loaded.

**Second way (cache policy)**
With Charles Proxy

1. In `DashboardViewController.swif` at line 21 replace to any image URL 
2. Click the `Start Video` button
3. Check if the requests were made
4. Back to the home screen
5. Click the `Start Video` button
6. Check if the requests were not made and the image was loaded


Sources/Clappr_iOS/Classes/Plugin/Container/PosterPlugin.swift
```diff
@@ -35,7 +33,7 @@ open class PosterPlugin: UIContainerPlugin {
     fileprivate typealias PosterUrl = String
     fileprivate func setPosterImage(with urlString: PosterUrl) {
         if let url = URL(string: urlString) {
-            poster.kf.setImage(with: url)
+            poster.download(url: url)
```
leandroalonso: What do you think about naming the method `display(url: URL)`? This is what the method does, right? It displays an image.

Sources/Clappr/Classes/Extension/UIImageView+Ext.swift
```diff
@@ -0,0 +1,24 @@
...
+                let image = UIImage(data: data)
+                else { return }
+
+            DispatchQueue.main.async() {
+                self.image = image
+            }
+        }.resume()
+    }
+
+    func download(string: String) {
```
leandroalonso: This method doesn't seem to be used anywhere. Should we keep it?

Sources/Clappr/Classes/Extension/UIImageView+Ext.swift
```diff
@@ -0,0 +1,24 @@
+public extension UIImageView {
+    func download(url: URL) {
+        let request = URLRequest(url: url, cachePolicy: .returnCacheDataElseLoad)
```
leandroalonso: Can we write a test for this method?

You could easily stub an image URL using `OHHttpStub` and then check if the image was correctly loaded.

Tests/Extensions/UIImageViewTests.swift
```diff
@@ -0,0 +1,79 @@
...
+                    var requestCount: Int!
+
+                    beforeEach {
+                        requestCount = 0
+
+                        stub(condition: isExtension("png") && isHost("test.io")) { _ in
+                            requestCount += 1
+                            let image = UIImage(named: "poster", in: Bundle(for: UIImageViewTests.self), compatibleWith: nil)!
+                            let data = UIImagePNGRepresentation(image)
+                            return OHHTTPStubsResponse(data: data!, statusCode: 200, headers: ["Content-Type":"image/jpeg"])
```
jademcosta: Minor comment/idea. As the image is a PNG, I believe the `Content-Type` should be `image/png`. Do you agree? Or it doesn't matter?

paulogama14: Make senses!

PR #279, opened 2019-02-15
Centralize Animation and Duration constants

caiobzen:
## 🏁 Goal
- To centralize all the animation and delays constants in one place
- To  make it easy to understand to what they refer to
- To make it possible to reuse the same constant in more than one place, so we can keep some of the animations synchronized

## ✅ How to test
As far as all the changes were intended to move these constants to another place, therefore not changing behavior, we believe that by running all the test, we should be good to go.


Sources/Clappr/Classes/Helper/AnimationDuration.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+public class AnimationDuration {
```
leandroalonso: `ClapprAnimationDuration` would make more sense in this case (and probably avoid any collision). What do you guys think?

PR #284, opened 2019-02-27
Update Player iOS Architecture

caiobzen:
## 🏁 Goal:
This PR aims to modify the current architecture to:
- Remove duplicated iOS and tvOS code
- Create non-visual Core and Container Plugins
- Move towards a similar architecture (comparing to Android and Web)

## ✅ How to test:
In order to test this Pull Request, you must:
- [ ] Make sure the tests pass
- [ ] Create a class that derives from CorePlugin or ContainerPlugin and make sure it can be loaded:
```    
class FakePlugin: CorePlugin {
    override static var name: String {
        return "FakeCorePlugin"
    }
    
    override var pluginName: String {
        return "FakeCorePlugin"
    }
    
    required init() {
        super.init()
    }
    
    @objc public required init(context: UIObject) {
        super.init()
        print("FakeCorePlugin is loaded!")
    }
}
```
- [ ] Add FakePlugin.self to the plugins list within `Player.swift`, line `214`
- [ ] Run the app and make sure it prints `FakeCorePlugin is loaded!` on the Xcode console


Sources/Clappr/Classes/Plugin/Container/UIContainerPlugin.swift
```diff
@@ -1,34 +1,15 @@
...
     }
-
-    @objc open func destroy() {
-        Logger.logDebug("destroying", scope: "UIContainerPlugin")
-        Logger.logDebug("destroying listeners", scope: "UIContainerPlugin")
-        stopListening()
-        Logger.logDebug("destroyed", scope: "UIContainerPlugin")
+    
+    open func render() {
+        //NSException(name: NSExceptionName("RenderNotOverriden"), reason: "UIContainerPlugins should always override the render method").raise()
```
geisonfgf: Can we remove this comment?

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
```diff
@@ -1,34 +1,15 @@
...
     }
-
-    @objc open func destroy() {
-        Logger.logDebug("destroying", scope: "UICorePlugin")
-        Logger.logDebug("destroying listeners", scope: "UICorePlugin")
-        stopListening()
-        Logger.logDebug("destroyed", scope: "UICorePlugin")
+    
+    open func render() {
+        //NSException(name: NSExceptionName("RenderNotOverriden"), reason: "UICorePlugins should always override the render method").raise()
```
geisonfgf: Can we remove this comment?

Sources/Clappr/Classes/Plugin/Container/UIContainerPlugin.swift
caiobzen: This is intended for the next step of the implementation.
There are tests set as `fit` to make sure every UIPlugin will override the render function.

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
caiobzen: Same as above! :D

PR #286, opened 2019-03-08
Handle all plugin types in Core and Container

leandroalonso:
# Goal

* To handle both `UIPlugin`'s and `Plugin`'s in `Core` and `Container`
* To force `UIPlugin`'s to implement `render()` method

# How to test

* Make sure that the player still behaves consistently


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/JumpMediaControlPlugin.swift
```diff
@@ -34,7 +34,9 @@ public class JumpMediaControlPlugin: JumpPlugin {
     
     private func filteredOutModalPlugins() -> [UICorePlugin]? {
         let pluginsWithoutMediaControl = core?.plugins.filter({ $0.pluginName != MediaControl.name })
-        return pluginsWithoutMediaControl?.filter({ ($0 as? MediaControlPlugin)?.panel != .modal })
+        return pluginsWithoutMediaControl?
+            .filter({ ($0 as? MediaControlPlugin)?.panel != .modal })
+            .compactMap({ $0 as? UICorePlugin })
```
igoriols: Hey @leandroalonso just remember to call `compactMap` before `filter`as we discussed, please.

PR #288, opened 2019-03-13
Swift 4.2 migration

leandroalonso:
# Goal

To migrate Clappr to Swift 4.2

# How to test

1. Make sure that app runs and video plays
2. Run tests


igoriols:
Hey @leandroalonso there are new versions of Quick/Nimble and OHTTPStubs that supports Swift 4.2.

PR #290, opened 2019-03-14
Trigger Quickseek event 

danielbarboza:
## Goal
To create new event:
- didTapQuickSeek: triggered when user double tap to seek 10 seconds.

## How to test
In iOS and tvOS: 
1. Open any vod;
2. Double tap on video
3. Check if didTapQuickSeek event were triggered;


dezouzart:
What do u think about making a test to ensure this event is been triggered when we use the feature?

danielbarboza:
> What do u think about making a test to ensure this event is been triggered when we use the feature?

done!

PR #292, opened 2019-03-15
Fix memory leak

igoriols:
# Goal

To fix memory leak issues.

When using `load` or `configure` before, the player ended up using so much memory and processing that the app was freezing.

# How to test

1. First, change the video that will be loaded. In `DashboardViewController.swift` change the option `kSourceUrl` to:

```
kSourceUrl: "https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_ts/master.m3u8"
```

2. In `ViewController.swift` change the `playing` listener to:

```
        player.on(Event.playing) { _ in
            print("on Play")
            DispatchQueue.main.asyncAfter(deadline: .now() + 4) {
                self.player.configure(options: self.options)
            }
        }
```

And `ready` listener to:

```
player.on(Event.ready) { _ in print("on Ready"); self.player.play() }
```

Load a video, grab a coffee, and just wait a couple of minutes. The app should not freeze and the video will be reloaded every 4 sec.

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/TimeIndicator.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/TimeIndicator.swift
@@ -67,7 +67,6 @@ open class TimeIndicator: MediaControlPlugin {
 
     required public init(context: UIObject) {
         super.init(context: context)
-        stopListening()
         bindEvents()
     }
 
@@ -76,6 +75,7 @@ open class TimeIndicator: MediaControlPlugin {
     }
 
     private func bindEvents() {
+        stopListening()
         bindContainerEvents()
         bindPlaybackEvents()
         bindCoreEvents()
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/TimeIndicator.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/TimeIndicator.swift
@@ -81,7 +81,7 @@ open class TimeIndicator: MediaControlPlugin {
         bindCoreEvents()
     }
 
-    private func bindContainerEvents() {
+    open func bindContainerEvents() {
         if let container = activeContainer {
             listenTo(container, eventName: Event.didChangePlayback.rawValue) { [weak self] (_: EventUserInfo) in self?.bindEvents() }
         }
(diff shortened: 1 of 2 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/TimeIndicator.swift
```diff
@@ -76,12 +75,13 @@ open class TimeIndicator: MediaControlPlugin {
...
 
     private func bindEvents() {
+        stopListening()
         bindContainerEvents()
         bindPlaybackEvents()
         bindCoreEvents()
     }
 
-    private func bindContainerEvents() {
+    open func bindContainerEvents() {
```
jademcosta: While reviewing, this change made me think about the "openess" of Clappr plugins. Maybe it's a discussion that is worth a small sync meeting. The question is: what will be the philosophy of all the built-in plugins in Clappr? Are we going to `open` all the methods from those plugins? 

I took a quick look at other plugins, and we don't have a standard on such methods. Some of them have private `bindXEvents` methods but some of them have half `open` and half `private`. I believe we should establish a standard on plugin methods. WDYT?

leandroalonso: @jademcosta this makes total sense.

I believe we can open another PR exposing the methods of our `MediaControl` plugins. That way every user can inherit those plugins if needed.

What do you think?

jademcosta: Makes sense :)

danielbarboza:
Hi Guys, 

The feature is ok, but I'd think if we need to stop listening to all events(core, container, and playback) when we change playback or change active container for example. 
It wouldn't be better only stop listening to playback events when the didChangePlayback event was triggered or stop listening to containers events when didChangeActiveContainer event was triggered.

leandroalonso:
@danielbarboza it makes sense.

But that would add a lot of code into the plugin: it will have to store the listener ids, then it will stop listening to these ids (and clearing them), then starting listening again (and storing the listener ids)...

Also, this is the same approach used in all the other plugins.

Can we have a discussion about it and - if necessary - change this behavior in other PR?

PR #293, opened 2019-03-26
Fix nowplaying on tvos

danielbarboza:
## Goal
Fix Now Playing to tvOS

## How to test
1 - run tests
2 - On `AVFoundationNowPlayingService` put a breakpoint at line `16` and check if items property is been set with video information(title, description, date, content identifier, watched time)

### DIFF
--- a/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
+++ b/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
@@ -1274,6 +1274,20 @@ class AVFoundationPlaybackTests: QuickSpec {
                         expect(nowPlayingService.countOfCallsOfSetItems).toEventually(equal(0))
                     }
                 }
+
+                context("when avplayer has playerItem and is ready to play") {
+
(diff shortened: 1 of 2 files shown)


Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -1274,6 +1274,20 @@ class AVFoundationPlaybackTests: QuickSpec {
                         expect(nowPlayingService.countOfCallsOfSetItems).toEventually(equal(0))
                     }
                 }
+
+                context("when avplayer has playerItem and is ready to play") {
+
+                    it("calls loadMetadata when the player is ready to play") {
```
caiobzen: Given the context, I think we can get rid of the redundancy here:
`it("calls loadMetadata when the player is ready to play")`
could be just
`it("calls loadMetadata")`

danielbarboza: done!

PR #295, opened 2019-04-02
Create media start time property on playback

dezouzart:
### Goal
To create media start time property on playback

### How to test
- run tests
- listen to `.playing` event and inspect `container.playback.mediaStartTime`


dezouzart:
playback already provide currentDate property

PR #299, opened 2019-04-17
didResize event

leandroalonso:
# Goal

To trigger a `didResize` event when player resizes.

# How to test

1. Subscribe to the container event `didResize`
2. Change the player size (by putting it in fullscreen or resizing the container view)
3. Check that the event is triggered

### DIFF
--- a/Sources/Clappr/Classes/Enum/Event.swift
+++ b/Sources/Clappr/Classes/Enum/Event.swift
@@ -53,4 +53,5 @@ public enum Event: String, CaseIterable {
     case didHideModal = "Clappr:didHideModal"
     case willShowModal = "Clappr:willShowModal"
     case willHideModal = "Clappr:willHideModal"
+    case didResize = "Clappr:didResize"
 }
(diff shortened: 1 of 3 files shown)


Sources/Clappr/Classes/Enum/Event.swift
```diff
@@ -53,4 +53,5 @@ public enum Event: String, CaseIterable {
     case didHideModal = "Clappr:didHideModal"
     case willShowModal = "Clappr:willShowModal"
     case willHideModal = "Clappr:willHideModal"
+    case didResize = "Clappr:didResize"
```
dezouzart: Do we want to make the player listen and trigger this event like it does with `requestFullscreen event`? it would make easier for apps to listen. `player.on(Event.didResize)`

dezouzart: and just reminding to include this new event in our doc = )

leandroalonso: Included.

For now, we will keep the `didResize` event only in Container.

PR #300, opened 2019-04-18
Add support to loop videos and bump minimum iOS support version to 10.0

caiobzen:
## 🏁 Goal
1. To be able to loop a video by using the option `kLoop`

## ✅ How to test
1. On `DashboardViewController.swift`, add the option `kLoop: true` to the list of options (line 23)
2. Open the app and play a video
3. When the video finishes, it should play again
4. Run the tests and make sure that they pass

* If you wish to test the "pre-buffering", set also the option `kStartAt: 10` and notice that once the video ends, there's no buffering happening when it plays again.

## 📱 Preview
kloop


Clappr.podspec
```diff
@@ -20,7 +20,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.ios.resources = 'Sources/Clappr_iOS/Resources/*.{xib,ttf,png,xcassets}'
 
-  s.ios.deployment_target = "9.0"
+  s.ios.deployment_target = "10.0"
```
rcavaltw: I've found some `if #available(iOS 10.0, *) {` in `QuickSeekPlugin`, `AVFoundationPlayback` and `AVFoundationPlayback+Bitrate`

Sources/Clappr/Classes/Extension/Dictionary+Ext.swift
```diff
@@ -14,3 +14,12 @@ extension Dictionary where Key == String, Value == Any {
         }
     }
 }
+
+public extension Dictionary where Key == String, Value == Any {
+    func bool(_ option: String) -> Bool {
```
rcavaltw:
```suggestion
    func bool(_ option: String, or alternative: Bool = false)
```
so that when using, we can make it explicit what happens in case the option is not present. For example, inside `AVFoundationPlayback.shouldLoop` you could write `options.bool(kLoop, or:false)`

caiobzen: I like that 👏

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -223,17 +225,27 @@ open class AVFoundationPlayback: Playback {
...
     }
-
+    
+    private var shouldLoop: Bool {
+        return options.bool(kLoop)
+    }
+    
+    private func newAVQueuePlayer(with item: AVPlayerItem) -> AVQueuePlayer {
+        let player = AVQueuePlayer(items: [item])
+        playerLooper = AVPlayerLooper(player: player, templateItem: item)
```
BruunoMM: I think this method's name leads to a misunderstanding of what it does. I mean, it not just instantiates a new `AVQueuePlayer` but it also sets the `playerLooper` variable.
In my opinion, the simplest solution would be to rename the method to something like `setupQueuePlayer` as it may include more steps than just creating a new instance of `AVQueuePlayer`. 

What do you guys think about that?

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -1119,6 +1119,19 @@ class AVFoundationPlaybackTests: QuickSpec {
                             expect(playback.isPlaying).to(beFalse())
                             expect(playback.isPaused).to(beFalse())
                         }
+                        
+                        context("and option kLoop is set to true") {
+                            it("plays the video again") {
+                                let options: Options = [kSourceUrl: "http://clappr.sample/master.m3u8", kLoop: true]
```
danielbarboza: Hi,
If I remove the kLoop option this test continues OK.

PR #302, opened 2019-04-29
Fix tvOS playback events

dezouzart:
### Problem
`willPlay` and `willPause` playback events are not been triggered on tvOS if users request it using the remote

### Goal
To trigger `willPlay` and `willPause` events on tvOS player

### How to test
1 - Check if those events are been triggered
2 - Run tests

### DIFF
--- a/Sources/Clappr_tvOS/Classes/Base/Player.swift
+++ b/Sources/Clappr_tvOS/Classes/Base/Player.swift
@@ -5,9 +5,10 @@ open class Player: UIViewController {
     fileprivate var playbackEventsListenIds: [String] = []
     fileprivate(set) var core: Core?
     static var hasAlreadyRegisteredPlaybacks = false
-    fileprivate var viewController: AVPlayerViewController?
+    fileprivate var viewController: AVPlayerViewController!
     private let baseObject = BaseObject()
     private var tvRemoteGesture: UITapGestureRecognizer?
+    private var tvRemoteHandler: TVRemoteHandler?
 
     override open func viewDidLoad() {
         core?.parentView = view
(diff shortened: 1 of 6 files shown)


Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -21,6 +22,8 @@ open class Player: UIViewController {
                 vc.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
                 view.addSubview(vc.view)
                 vc.didMove(toParent: self)
+                
+                addTvRemoveGestureIfNeeded(vc)
```
paulogama14: I think this method has more than one responsibility, maybe we can extract the check to another method.

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -29,6 +32,30 @@ open class Player: UIViewController {
 
         core?.render()
     }
+    
+    func addTvRemoveGestureIfNeeded(_ viewController: AVPlayerViewController) {
```
paulogama14: There is a typo on the method name.

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -29,6 +32,30 @@ open class Player: UIViewController {
...
+        if let tvRemoteGesture = tvRemoteGesture,
+            let viewGestures = viewController.view.gestureRecognizers,
+            viewGestures.contains(tvRemoteGesture) {
+            return
```
paulogama14: Why have we to check this?

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -29,6 +32,30 @@ open class Player: UIViewController {
...
+        gestureRecognizer.allowedPressTypes = [NSNumber(value: UIPress.PressType.playPause.rawValue),
+                                               NSNumber(value: UIPress.PressType.select.rawValue)]
+        
+        viewController.view.addGestureRecognizer(gestureRecognizer)
+        self.tvRemoteGesture = gestureRecognizer
+    }
+
+    @objc func handleTvRemoteGesture() {
+        guard let playback = activePlayback as? AVFoundationPlayback else { return }
+        if playback.isPaused {
```
paulogama14: I think that we have to respect the user action, not infer based on `playback` state. I mean, if we're trying to make a pause we should call a pause. 
Make sense?

leandroalonso: The remote controller has just one button to play/pause and a select button that acts the same. This will respect the user action.

PR #303, opened 2019-05-01
Add a way to listenTo events without having to pass in the event rawValue

caiobzen:
## 🏁 Goal
- To reduce the "verbosity on listenTo events
- Remove the public keyword within a public extension to remove redundancy

## ✅ How to test
1. On any listenTo events, try to replace it with event: .eventName instead. Is should still work

### DIFF
--- a/Sources/Clappr/Classes/Base/BaseObject.swift
+++ b/Sources/Clappr/Classes/Base/BaseObject.swift
@@ -129,11 +129,15 @@ open class BaseObject: NSObject, EventProtocol {
 }
 
 public extension BaseObject {
-    public func trigger(_ event: Event) {
+    func trigger(_ event: Event) {
         trigger(event.rawValue)
     }
 
-    public func trigger(_ event: Event, userInfo: EventUserInfo) {
+    func trigger(_ event: Event, userInfo: EventUserInfo) {
         trigger(event.rawValue, userInfo: userInfo)
+    }
+
+    func listenTo<T: EventProtocol>(_ contextObject: T, event: Event, callback: @escaping EventCallback) {
+        listenTo(contextObject, eventName: event.rawValue, callback: callback)
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/Clappr/Classes/Base/BaseObject.swift
```diff
@@ -129,11 +129,15 @@ open class BaseObject: NSObject, EventProtocol {
 }
 
 public extension BaseObject {
```
rcavaltw: Although this is an extension to `BaseObject`, I would move it to `Event.swift` as it is more about `Event` than about `BaseObject`. Thoughts?

Sources/Clappr/Classes/Base/BaseObject.swift
```diff
@@ -129,11 +129,15 @@ open class BaseObject: NSObject, EventProtocol {
...
+    func trigger(_ event: Event) {
         trigger(event.rawValue)
     }
 
-    public func trigger(_ event: Event, userInfo: EventUserInfo) {
+    func trigger(_ event: Event, userInfo: EventUserInfo) {
         trigger(event.rawValue, userInfo: userInfo)
     }
+
+    func listenTo<T: EventProtocol>(_ contextObject: T, event: Event, callback: @escaping EventCallback) {
```
igoriols: Are we going to start replacing the calls on demand?

jademcosta: What would you do, @igoriols ? :P I confess I have doubts. Maybe we can convert all of them on a separate PR?

Sources/Clappr/Classes/Base/BaseObject.swift
caiobzen: That’s a good point. But I don’t think this is a reason not to be close to BaseObject. For me, it feels like it should be here cause I’m extending this class behavior

PR #304, opened 2019-05-03
Plugin.pluginName refactor

rcavaltw:
# Goal
Reduce the startup time related to instantiating plugins unnecessarily. The current implementation of `Plugin.name`, requires a call to the instance method `pluginName`, therefore instantiating plugins unnecessarily during the `Loader.register` flow.

The meat of this refactoring is inside `Plugin.swift`, `CorePlugin.swift` and `ContainerPlugin.swift`, where the implementations of `name` and `pluginName` were changed.

As a side effect, the parameterless `init()` method was not required anymore and was also removed from all Plugins.

# How to test
* `make test` and  Run the example application.

### DIFF
--- a/Sources/Clappr/Classes/Protocol/Plugin.swift
+++ b/Sources/Clappr/Classes/Protocol/Plugin.swift
@@ -12,7 +12,7 @@ public func theType<T>(of value: T) -> Plugin.Type? {
 
 extension Plugin {
 
-    var pluginName: String {
+    public var pluginName: String {
         guard let selfType = theType(of: self) else { return "" }
         return selfType.name
     }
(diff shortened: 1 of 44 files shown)


Sources/Clappr/Classes/Protocol/Plugin.swift
```diff
@@ -1,8 +1,15 @@
...
+public protocol Plugin: EventProtocol, Nameable {
     static var type: PluginType { get }
-    static var name: String { get }
-    var pluginName: String { get }
-    init()
     init(context: UIObject)
     func destroy()
 }
+
+public protocol Nameable {
```
fcapinho: This should be named `NamedType` to be consistent with Android implementation.

igoriols: The presence of the word **Type** in the name of a protocol/interface looks strange since it already represents a type. It looks a little redundant. Can we keep the name `Nameable` and change in Android in the future?

fcapinho: No, we should keep the consistency and reopen the discussion in the future.

PR #306, opened 2019-05-13
Bind events refactor

rcavaltw:
# Goal

The objective of this refactor is push the `bindEvents` up in the hierarchy of plugins so that plugins don't forget to `stopListening` or to re-`bindEvents` when `didChangeActiveContainer` and `didChangeActivePlayback` are triggered.

The main changes are in `ContainerPlugin.swift` and `CorePlugin.swift`. All other changes are so that plugins comply with the new structure.

A secondary benefit is that many plugins now don't need to override the required constructor when the only thing it used to do was to bind events.

# How to test
1. Run `make test`
2. Test the basic functionality in the example app.

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Core/CorePlugin.swift
+++ b/Sources/Clappr/Classes/Plugin/Core/CorePlugin.swift
@@ -30,13 +30,12 @@ open class CorePlugin: BaseObject, Plugin {
             self?.bindAllEvents()
             self?.onDidChangePlayback()
         }
-
-
     }
 
     open func bindEvents() { }
 
     open func onDidChangePlayback() { }
+
     open func onDidChangeActiveContainer() { }
 
     open func destroy() {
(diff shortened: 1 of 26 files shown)


Sources/Clappr/Classes/Plugin/Core/CorePlugin.swift
```diff
@@ -14,8 +14,31 @@ open class CorePlugin: BaseObject, Plugin {
...
+        listenTo(core, eventName: Event.didChangeActiveContainer.rawValue) { [weak self] _ in
+            self?.bindAllEvents()
+            self?.onDidChangePlayback()
+        }
+        guard let container = core.activeContainer else { return }
+        listenTo(container, event: .didChangePlayback) { [weak self] _ in
+            self?.bindAllEvents()
+            self?.onDidChangePlayback()
+        }
+
```
caiobzen: I think we could remove the extra lines here

PR #317, opened 2019-06-06
Trigger didLoop event for every video loop

caiobzen:
## 🏁 Goal
- To trigger a `didLoop` event on every video loop

## ✅ How to test
1. On `DashboardViewController.swift`, add a new option: `kLoop: true`
2. On AVFoundationPlayback.swift, place a breakpoint on line 25, within the observer scope
3. Play the video and drag the scrubber to the end of the video
4. You should see the code stopping on your breakpoint
5. You can repeat steps 3 and 4 just to make sure it is always triggering the `didLoop` event.

### DIFF
--- a/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
+++ b/Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
@@ -86,6 +86,18 @@ class AVFoundationPlaybackTests: QuickSpec {
                             expect(didLoopTriggered).toEventually(beTrue(), timeout: 3)
                         }
                     }
+
+                    context("on stop") {
+                        it("stops observing loopCount") {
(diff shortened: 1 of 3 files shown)


Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -68,6 +68,24 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                                didLoopTriggered = true
+                            }
+
+                            playback.play()
+                            playback.seek(playback.duration)
+
+                            expect(didLoopTriggered).toEventually(beTrue(), timeout: 3)
+                        }
+                    }
                 }
```
paulogama14: Could we create a test setting looperObserver to nil?

PR #318, opened 2019-06-06
Expose wait in playback

rcavaltw:
# Goal
Provide to plugins the ability to wait for `Playback` properties to be accessed after the dependent `AVAsset` properties are properly loaded

# How to test
Try to use the `Playback.wait(for:completion:)` method in the sample app.


towerz:
@rcavaltw aren't you attempting to expose too much of the internal Playback implementation with this? 

The problem with that is that it creates as many "intermediary states" for the Playback as the properties that you want to check whether they're available.

I believe this could solve some internal issues, but it shouldn't be exposed to Plugins. Instead, the availability of some properties should correspond to events externally. This way it keeps communication between the Playback and the plugins using a single method, through event listeners. Having `wait` exposed adds a listener-based (or delegate-like) communication between the Playback and the plugins.

Another problem is that if there is another Playback implementation (for example a low level, VLC-like implementation), we wouldn't be able to guarantee the same behavior. This actually creates a strong dependency on AVPlayer (and related classes), which we shouldn't pursue.

rcavaltw:
A future card will address proper `.ready` semantics, making this PR unecessary. Closing it for now.

PR #322, opened 2019-06-28
Expose FocusEnvironments for tvOS

caiobzen:
## 🏁 Goal:
- To make it possible to have focusable UI elements over the `tvOS` player, so we can customize it in the future.

## ✅ How to test:
1. On tvOS `Player.swift`, create a dummy button and add it to the contentOverlayView as a subView:
```swift
let button = UIButton(type: .system)
button.frame = CGRect(x:0, y:0, width: 100, height: 100)
button.backgroundColor = .red
contentOverlayView?.addSubview(button)
```

2. On `ViewController.swift`, add the following code:
```swift
open override var preferredFocusEnvironments: [UIFocusEnvironment] {
    return player.focusEnvironments
}
```

3. Once you run the app, make sure the button is focused and you can interact with it.


Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -48,7 +46,7 @@ open class Player: UIViewController {
     }
 
     open var isFullscreen: Bool {
-        guard let core = self.core else {
+        guard let core = core else {
```
rcavaltw: Since you touched, you might as well make it a one liner. 
```suggestion
        core?.isFullscreen ?? false
```
On a side note, I wonder if for tvos, core.isFullscreen should default to `true` 🤔

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -289,3 +274,31 @@ open class Player: UIViewController {
...
+extension Player {
+    private var isPaused: Bool {
+        return activePlayback?.state == .paused
+    }
+
+    private var isPlaying: Bool {
+        return activePlayback?.state == .playing
+    }
+
+    open override func pressesBegan(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
```
dezouzart: An override func in a `extension class` feels strange, don't you think? What about making this extension be like `extension class: superClass`  or moving this func to the class implementation?

caiobzen: The main reason for doing such a thing was to keep the context of the deleted class `DecoratedAVPlayerViewController`. This code is not meant to be used within the Player itself, so that's the main reason to keep it separate in an extension. 

What do you think?

dezouzart: I usually avoid to create extensions without a protocol to conform, like `extension OwnMyClass`, I move funcs to the class itself. But we don't have a team agreement for this, so it's up 2 u. = )

PR #325, opened 2019-07-03
Select off subtitles

rcavaltw:
# Goal
Fix a problem in tvOS where the user cannot turn subtitles off after choosing a language.

# Test

Check manually, I've already changed the example tvos video for a bip-bop with subtitles!


Sources/Clappr/Classes/Base/MediaOption.swift
```diff
@@ -1,10 +1,10 @@
 import Foundation
```
BruunoMM: I think we can remove this import. 🙋‍♂

PR #326, opened 2019-07-08
Expose FocusEnvironments for tvOS

igoriols:
## 🏁 Goal:
- To make it possible to have focusable UI elements over the `tvOS` player, so we can customize it in the future.

## ✅ How to test:
1. On tvOS `Player.swift`, create a dummy plugin at the bottom of the file:
```swift
class ButtonPlugin: UICorePlugin {
    override open class var name: String {
        return "ButtonPlugin"
    }

    override func bindEvents() { }

    override func render() {
        let button = UIButton(type: .system)
        button.frame = CGRect(x: 0, y: 0, width: 100, height: 100)
        button.backgroundColor = .red
        core?.view.addSubview(button)
    }
}
```

2. Register it on `Player:L102`:
```swift
Player.register(plugins: [ButtonPlugin.self])
 ```

3. On `ViewController.swift`, add the following code:
```swift
open override var preferredFocusEnvironments: [UIFocusEnvironment] {
    return player.focusEnvironments
}
```

4. Once you run the app, make sure the button is focused and you can interact with it.


igoriols:
[Pular abertura/resumo para tvOS](https://trello.com/c/EflY6Z6d/617-pular-abertura-resumo-para-tvos)

PR #329, opened 2019-07-13
Trigger willLoadSource from core when player.configure or load are called

rcaval:
# Goal

To trigger `.willLoadSOurce` from core whenever `player.configure` or player.load is called. This should let plugins prepare for fresh options if they need to.

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -176,6 +176,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
 
     fileprivate func loadSourceIfNeeded() {
         if let source = options[kSourceUrl] as? String {
+            trigger(.willLoadSource)
             activeContainer?.load(source, mimeType: options[kMimeType] as? String)
         }
     }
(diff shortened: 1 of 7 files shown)


Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -176,6 +176,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
 
     fileprivate func loadSourceIfNeeded() {
         if let source = options[kSourceUrl] as? String {
+            trigger(.willLoadSource)
```
fcapinho: Recommend adding a `load` method to `Core` to separate it from options change. `load` should not be an options update side-effect.

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -142,6 +142,7 @@ open class Player: AVPlayerViewController {
     }
 
     open func load(_ source: String, mimeType: String? = nil) {
+        core?.trigger(.willLoadSource)
         activeContainer?.load(source, mimeType: mimeType)
```
fcapinho: `Player` should call `Core.load` instead of 'Core.activeContainer.load`

Sources/Clappr/Classes/Base/Core.swift
igoriols: Done.

Sources/Clappr_tvOS/Classes/Base/Player.swift
igoriols: Done.

PR #331, opened 2019-07-17
Create didChangeScreenOrientation and didDoubleTouchMediaControl events 

paulogama14:
## 🏁 Goal:
- To trigger a `didChangeScreenOrientation` event every orientation change.
- To trigger a `didDoubleTouchMediaControl` event every consecutive double tap on screen.

## ✅ How to test `didChangeScreenOrientation`:
1. On `ViewController.swift#52` add this piece of code:
```swift
player.on(Event.didChangeScreenOrientation) { _ in
    print("on didChangeScreenOrientation")
}
```
2. On `Player.swift#100` add the event to trigger on player layer:
```swift
core?.activeContainer?.on(Event.didChangeScreenOrientation.rawValue) { [weak self] (info: EventUserInfo) in self?.trigger(Event.didChangeScreenOrientation.rawValue, userInfo: info) }
```
3. Check the console log and ensure that `on didChangeScreenOrientation` was printed

## ✅ How to test `didDoubleTouchMediaControl`:
1. On `ViewController.swift#52` add this piece of code:
```swift
player.on(Event.didDoubleTouchMediaControl) { _ in
    print("on didDoubleTouchMediaControl")
}
```
2. On `Player.swift#100` add the event to trigger on player layer:
```swift
core?.activeContainer?.on(Event.didDoubleTouchMediaControl.rawValue) { [weak self] (info: EventUserInfo) in self?.trigger(Event.didDoubleTouchMediaControl.rawValue, userInfo: info) }
```
3. Check the console log and ensure that `on on didChangeScreenOrientation with params -> Optional([AnyHashable("position"): "right"])` and `on didChangeScreenOrientation with params -> Optional([AnyHashable("position"): "left"])` was printed with `right` and `left` according to tap screen position

### DIFF
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -32,9 +32,7 @@ open class Container: UIObject {
     }
 
     private var boundsObservation: NSKeyValueObservation?
-    #if os(iOS)
-    private var previousDeviceOrientation: UIDeviceOrientation = UIDevice.current.orientation
-    #endif
+
     public init(options: Options = [:]) {
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
         self.options = options
@@ -73,9 +71,6 @@ open class Container: UIObject {
         renderPlayback()
 
         observeWhenViewChangeBounds()
-        #if os(iOS)
-        addOrientationObserver()
-        #endif
     }
 
     fileprivate func renderPlayback() {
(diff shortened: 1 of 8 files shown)


Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -118,6 +123,32 @@ open class Container: UIObject {
         }
     }
 
+    #if os(iOS)
```
caiobzen: I tend to re-think about having the container handling such responsibility.

having platform checks like`#if os(iOS)` may suggest we want to move all this orientation thing to another class.

What do you think about having a `DeviceOrientationPlugin` responsible for listening to  `UIDevice.orientationDidChangeNotification` and triggering the device rotation events?

This way we can add this plugin to the iOS Player plugins list (since `tvOS` will not take advantage of it)

paulogama14: Nice

=]

PR #332, opened 2019-07-22
Add sharedData to core

caiobzen:
## 🏁 Goal:
- To add to Core the ability to store shared data across the player lifecycle

## ✅ How to test:
1. Store anything you want on `core.sharedData` and try to retrieve it
2. Run the tests (they execute the step 1).

### DIFF
--- a/Tests/Clappr_Tests/CoreTests.swift
+++ b/Tests/Clappr_Tests/CoreTests.swift
@@ -96,8 +96,15 @@ class CoreTests: QuickSpec {
             }
 
             describe("Core sharedData") {
-                context("when stores a value on sharedData") {
+                context("on a brand new instance") {
+                    it("starts empty") {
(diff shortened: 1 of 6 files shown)


Tests/Clappr_Tests/CoreTests.swift
```diff
@@ -95,6 +95,20 @@ class CoreTests: QuickSpec {
...
             }
 
+            describe("Core sharedData") {
+                context("when stores a value on sharedData") {
+
+                    beforeEach {
+                        core = CoreFactory.create(with: [:])
+                        core.sharedData["testKey"] = "testValue"
+                    }
+
```
paulogama14: Could we create a test to ensure that the initial value is empty and after erase content the shared data was cleared?

caiobzen: I did the same with the container tests! :)

PR #334, opened 2019-07-25
Change way to trigger willPause event in tvOS

igoriols:
### Goal
Willpause event was being triggered in `Player` for tvOS, but it was not working on a real device.

We had to change where the event is triggered. The event is being triggered in `AVFoundationPlayback`. 

### How to test
Run tvOS sample and check if `on willPause` is printed when the video is paused.

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -370,7 +370,7 @@ open class AVFoundationPlayback: Playback {
 
     private func handlePlayerRateChanged(_ player: AVPlayer, _ changes: NSKeyValueObservedChange<Float>) {
         guard playerStatus != .unknown else { return }
-        if changes.isPrior && player.rate == 1 && state == .playing {
+        if changes.isPrior && player.rate != 0.0 && state == .playing {
             triggerWillPause()
         } else if !changes.isPrior && player.rate == 0 && state != .idle {
             updateState(.paused)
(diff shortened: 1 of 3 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -356,12 +357,18 @@ open class AVFoundationPlayback: Playback {
...
+            Logger.logError("It was not possible to set the audio session category")
         }
     }
-
-    private func handlePlayerRateChanged(player: AVPlayer, changes: Any) {
-        if player.rate == 0 && playerStatus != .unknown && state != .idle {
+    
+    private func handlePlayerRateChanged(player: AVPlayer, changes: NSKeyValueObservedChange<Float>) {
+        guard playerStatus != .unknown else { return }
+        if changes.isPrior && player.rate == 1 && state == .playing {
```
lfonseca: O teste nesse condicional não seria `player.rate > 0`?
Teoricamente qualquer valor maior que zero indica que o state do player é `playing`.

rcavaltw: For some reason I don't remember anymore, we made it playing only when `rate == 1`. I've move it to `rate != 0` again and in other place too. Nothing breaks, so I guess we are good to go :)

lfonseca: Great! GTG to me as well :)

PR #335, opened 2019-08-06
Trigger core ready event after rendering containers

caiobzen:
## 🏁  Goal:
1. In order to know when `core` is fully loaded (regardless of the playback state), we are conforming to the `Clappr web` approach that is to trigger a `CORE_READY` event right after containers are rendered.

## ✅  How to test:
1.  Create a dummy plugin that listens to .coreReady
2. Load the plugin in the plugins list
3. Launch the sample app
4. Notice that your plugin will listen to the core ready event right away

### Dummy plugin example:
```swift
class DummyPlugin: SimpleCorePlugin {
        return "DummyPlugin"
    }

    override func bindEvents() {
        listenTo(core!, event: .coreReady) { _ in
            print("I'm listening to coreReady event!")
        }
    }
}
```

Just add it to the player `builtInPlugins` list

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -115,6 +115,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     open override func render() {
         containers.forEach(renderContainer)
         addToContainer()
+        trigger(.coreReady)
     }
 
     #if os(tvOS)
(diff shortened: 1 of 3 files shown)


Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -115,6 +115,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     open override func render() {
         containers.forEach(renderContainer)
         addToContainer()
+        trigger(.coreReady)
```
rcavaltw: just wondering, why not use the existing `.ready` event? As it is triggered in core, it is already bound to it.

caiobzen: Because `.ready` is being used by AVFoundationPlayback only. We decided to use the same event that Clappr web is using, which is `coreReady`.
And I'm not seeing `.ready` being fired of by core 🤔

PR #336, opened 2019-08-06
Create didUpdateBitrate event

igoriols:
### Goal

Create `didUpdateBitrate`event.

### How to test

1. Add those lines to `ViewController` inside method `listenToPlayerEvents`:
```swift
player.on(Event.didUpdateBitrate) { userInfo in print("on didUpdateBitrate with: \(userInfo?["bitrate"])") }
```
2. Run sample app and check if the message is printed.

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -296,15 +296,9 @@ open class AVFoundationPlayback: Playback {
     }
     
     @objc func onAccessLogEntry(notification: NSNotification?) {
-        guard let playerItem = notification?.object as? AVPlayerItem,
-            let lastEvent = playerItem.accessLog()?.events.last else {
-                return
-        }
-        
-        if lastBitrate != lastEvent.bitrate {
-            lastBitrate = lastEvent.bitrate
-            trigger(.didUpdateBitrate, userInfo: ["bitrate": lastBitrate ?? 0.0])
-        }
+        guard lastBitrate != bitrate else { return }
+        lastBitrate = bitrate
+        trigger(.didUpdateBitrate, userInfo: ["bitrate": lastBitrate ?? 0.0])
     }
 
     private var hasEnoughBufferToPlay: Bool {
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -249,7 +249,7 @@ open class AVFoundationPlayback: Playback {
         playerLayer?.frame = view.bounds
         setupMaxResolution(for: playerLayer!.frame.size)
 
-        asset.wait(for: .characteristics, then: selectDefaultAudioIfNeeded)
+        asset.wait(for: .characteristics, then: selectDefaultMediaOptionIfNeeded)
         asset.wait(for: .duration, then: seekToStartAtIfNeeded)
         addObservers()
     }
@@ -314,7 +314,6 @@ open class AVFoundationPlayback: Playback {
 
         if hasEnoughBufferToPlay {
             play()
-            selectDefaultSubtitleIfNeeded()
         } else {
             updateState(.stalling)
         }
@@ -558,6 +557,11 @@ open class AVFoundationPlayback: Playback {
         return source.first(where: { $0.language == language })?.raw as? AVMediaSelectionOption
     }
 
+    func selectDefaultMediaOptionIfNeeded() {
+        selectDefaultAudioIfNeeded()
+        selectDefaultSubtitleIfNeeded()
+    }
+    
     internal func selectDefaultSubtitleIfNeeded() {
         guard let subtitles = subtitles else { return }
         var isFirstSelection = false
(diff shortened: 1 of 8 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -286,6 +287,24 @@ open class AVFoundationPlayback: Playback {
             selector: #selector(AVFoundationPlayback.playbackDidEnd),
             name: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
             object: player.currentItem)
+        
+        NotificationCenter.default.addObserver(
```
caiobzen: Since bitrate is an "agnostic" information regardless of the playback type, does it make sense to move the bitrate tracking responsibility out of the `AVFoundationPlayback`? 

The benefits I can point out so far:
1. One less responsibility for the AVFoundationPlayback class
2. `.didUpdateBitrate` event wouldn't be coupled to `AVFoundationPlayback`
3. Other playbacks could benefit from this bitrate event as well

I was thinking if the `Playback` class could play this role instead. 
The only thing we should create an abstraction for is the notification mechanism.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -286,6 +287,24 @@ open class AVFoundationPlayback: Playback {
...
+            self,
+            selector: #selector(AVFoundationPlayback.onAccessLogEntry),
+            name: NSNotification.Name.AVPlayerItemNewAccessLogEntry,
+            object: nil)
+    }
+    
+    @objc func onAccessLogEntry(notification: NSNotification?) {
```
caiobzen: I think we can create an extension of AVPlayerItem to encapsulate this code:
```swift
extension AVPlayerItem {
    var bitrate: Double? {
        return accessLog()?.events.last?.bitrate ?? 0.0
    }
}
```

ending up with a simpler guard let statement:
```swift
    @objc func onAccessLogEntry(notification: NSNotification?) {
        guard let playerItem = notification?.object as? AVPlayerItem,
            lastBitrate != playerItem.bitrate else { return }

        lastBitrate = playerItem.bitrate
        trigger(.didUpdateBitrate, userInfo: ["bitrate": lastBitrate])
    }
```

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -507,7 +523,7 @@ open class AVFoundationPlayback: Playback {
     }
 
     func handleDvrAvailabilityChange() {
-        if lastDvrAvailability != isDvrAvailable {
+        if lastDvrAvailability != isDvrAvailable && playbackType == .live {
```
caiobzen: Could you guys please make this if statement more readable? 
something like:
```swift
private var dvrAvailabilityChanged: Bool {
    return lastDvrAvailability != isDvrAvailable && playbackType == .live
}

func handleDvrAvailabilityChange() {
    guard dvrAvailabilityChanged else { return }
    trigger(.didChangeDvrAvailability, userInfo: ["available": isDvrAvailable])
    lastDvrAvailability = isDvrAvailable
}
```

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -507,12 +524,16 @@ open class AVFoundationPlayback: Playback {
...
     func handleDvrAvailabilityChange() {
-        if lastDvrAvailability != isDvrAvailable {
+        if dvrAvailabilityChanged {
             trigger(.didChangeDvrAvailability, userInfo: ["available": isDvrAvailable])
             lastDvrAvailability = isDvrAvailable
         }
     }
 
+    private var dvrAvailabilityChanged: Bool {
+        return lastDvrAvailability != isDvrAvailable && playbackType == .live
```
paulogama14: I think that we can change the order because the playbackType is more important than DVR availability.

igoriols: Done.

Tests/Clappr_Tests/AVFoundationPlaybackTests.swift
```diff
@@ -240,9 +257,9 @@ class AVFoundationPlaybackTests: QuickSpec {
                     var didCallChangeDvrAvailability: Bool?
                     let playerAsset = AVURLAssetStub(url: URL(string: "http://clappr.sample/master.m3u8")!)
                     
-                    func setupTest(minDvrSize: Double, seekableTimeRange: Double) {
+                    func setupTest(minDvrSize: Double, seekableTimeRange: Double, duration: CMTime = CMTime.indefinite) {
```
caiobzen: I think you can remove the prefix of `CMTime.indefinite`, leaving only the `.indefinite`.

caiobzen: func setupTest(minDvrSize: Double, seekableTimeRange: Double, duration: CMTime = .indefinite) {

igoriols: Done.

PR #340, opened 2019-08-22
Overlay plugin

caiobzen:
## 🏁  Goal:
We are introducing a new type of plugin which the main goal is to render over the top of everything.
Overlay Plugins can be useful when you have to display something over the whole content, such as a settings panels, ads, etc.

## ✅  How to test:

1. Go to `Player.swift` file and at the very bottom, paste the following code:

```swift
class MyOverlay: OverlayPlugin {
    override class var name: String {
        return "MyOverlay"
    }

    override func render() {
        let overlayView = UIVisualEffectView(effect: UIBlurEffect(style: .regular))
        overlayView.frame = CGRect(x: 20, y: 20, width: 200, height: 60)
        overlayView.layer.cornerRadius = 5.0
        overlayView.clipsToBounds = true

        let label = UILabel(frame: overlayView.frame)
        label.text = "I'm an overlay plugin!"
        label.textAlignment = .center
        label.textColor = .white
        label.shadowColor = .black
        label.font = .boldSystemFont(ofSize: 16)
        label.shadowOffset = CGSize(width: 1, height: 1)

        view.addSubview(overlayView)
        view.addSubview(label)
    }
}
```

2. Add this plugin to the `builtInPlugins` list:
Captura de Tela 2019-08-22 às 10 25 59

3. Run the app and you should see something like the screenshot below:
Captura de Tela 2019-08-22 às 10 27 28


Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -219,14 +223,46 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
             try ObjC.catchException {
-                plugin.destroy()
+                self.destroy()
+            }
+        } catch {
+            logDestroyCrash(for: error)
+        }
+    }
+
+    func safeRender() {
```
lfonseca: Once you have typecasted `plugin` before using it, wouldn't it make more sense to add this method to a `UICorePlugin` extension?

PR #345, opened 2019-09-11
DrawerPlugin

caiobzen:
## Goal 🏁 
- To introduce a base class for creating drawer plugins (plugins that will be placed at the edges of the player, like floating panels)
- We haven't added the possible drawer positions yet as they might be added as we develop the drawers. Therefore, the only position available for now is `undefined`

## How to test ✅ 
- Run the test suite and nothing should break
- You can create a drawer plugin to test out the behaviors:

```swift
// Add this plugin to the built in plugins list!
class MyDrawer: DrawerPlugin {
    override class var name: String {
        return "MyDrawer"
    }

    override var position: DrawerPlugin.Position {
        return .undefined
    }

    override var size: CGSize {
        return CGSize(width: 200, height: 100)
    }

    let toggleButton = UIButton(frame: CGRect(origin: .zero, size: CGSize(width: 200, height: 100)))

    override func render() {
        super.render()
        view.frame = CGRect(origin: CGPoint(x: 20, y: 20), size: size)

        toggleButton.backgroundColor = .init(white: 0.9, alpha: 0.9)
        toggleButton.tintColor = .black
        toggleButton.addTarget(self, action: #selector(toggleDrawer), for: .touchUpInside)

        updateText()

        view.addSubview(toggleButton)
    }

    private func updateText() {
        toggleButton.setTitle("isClosed: (\(isClosed))", for: .normal)
    }

    @objc func toggleDrawer() {
        let event: Event = isClosed ? .showDrawerPlugin : .hideDrawerPlugin
        core?.trigger(event)
        updateText()
    }
}
```

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
@@ -37,6 +37,19 @@ class DrawerPlugin: OverlayPlugin {
     override func bindEvents() {
         guard let core = core else { return }
 
+        listenTo(core, event: .willShowMediaControl) { [weak self] _ in
+            UIView.animate(withDuration: ClapprAnimationDuration.mediaControlShow) {
+                self?.view.alpha = 1
+            }
+        }
+
+        listenTo(core, event: .willHideMediaControl) { [weak self] _ in
+            guard self?.isClosed == true else { return }
+            UIView.animate(withDuration: ClapprAnimationDuration.mediaControlHide) {
+                self?.view.alpha = 0
+            }
+        }
+
         listenTo(core, event: .showDrawerPlugin) { [weak self] _ in
             guard self?.isClosed != false else { return }
             self?.isClosed = false
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
@@ -57,14 +57,17 @@ open class DrawerPlugin: OverlayPlugin {
         guard let core = core else { return }
 
         listenTo(core, event: .showDrawerPlugin) { [weak self] _ in
-            guard self?.isClosed != false else { return }
-            self?.isClosed = false
+            self?.toggleIsClosed(to: false)
         }
 
         listenTo(core, event: .hideDrawerPlugin) { [weak self] _ in
-            guard self?.isClosed != true else { return }
-            self?.isClosed = true
+            self?.toggleIsClosed(to: true)
         }
+    }
+
+    private func toggleIsClosed(to newValue: Bool) {
+        guard isClosed != newValue else { return }
+        isClosed = newValue
     }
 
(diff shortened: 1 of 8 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
```diff
@@ -0,0 +1,82 @@
...
+                self?.view.alpha = 0
+            }
+        }
+    }
+
+    private func bindDrawerEvents() {
+        guard let core = core else { return }
+
+        listenTo(core, event: .showDrawerPlugin) { [weak self] _ in
+            self?.toggleIsClosed(to: false)
```
dezouzart: why not call `self?.showDrawer()` or something like this?

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
```diff
@@ -0,0 +1,82 @@
...
+        bindMediaControlEvents()
+        bindDrawerEvents()
+    }
+
+    private func bindMediaControlEvents() {
+        guard let core = core else { return }
+
+        listenTo(core, event: .willShowMediaControl) { [weak self] _ in
+            UIView.animate(withDuration: ClapprAnimationDuration.mediaControlShow) {
+                self?.view.alpha = 1
```
dezouzart: Do we have to check if drawer is open before make it visible?

dezouzart:
`make test` command is failling on this branch =/

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
caiobzen: There's a reason:

## The current solution:
- `.showDrawerPlugin` event is going to toggle the isClosed flag to `false` _(if it's not false yet)_
- `.hideDrawerPlugin` event is going to toggle the isClosed flag to `true` _(if it's not true yet)_

The function `toggleIsClosed` is handling both cases.

## The proposed change
If we go with the suggested approach, we gonna end up with 2 functions:
- `showDrawer`
- `hideDrawer` 

Both functions are going to duplicate the check if the Bool is already `true` or `false`. 
We've tried this approach before and `CodeClimate` was complaining about it. 


## Conclusion
The Idea was to keep a self explanatory name that reads like an instruction.
if the wording here `toggle "is closed" to false` is not explicit enough, we can come up with a better name for it.

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
caiobzen: The `isClosed` state stands for wether it should keep visible if the mediaControl is going to be **dismissed**. In this case, we always want the drawer to be visible when mediaControl is going to be presented. Therefore, we don't have to implement such verification.

caiobzen:
`make test` is up and running again ✅

caiobzen:
Updated events documentation:
updated docs

PR #347, opened 2019-09-18
Bottom drawer plugin

paulogama14:
## Goal

We are introducing the BottomDrawerPlugin structure, that allows customizing `DrawerPlugin` with a bottom behavior. With this, it's possible to create a plugin from the bottom screen.

Simulator Screen Shot - iPhone 8 - 2019-09-18 at 18 01 29

## How to test

1 - Add this code on BottomDrawerPlugin.swift#127 

```swift
    }

        return 32.0
    }

    override func render() {
        super.render()

        view.backgroundColor = .red
    }
}
```
2 - Add `CustomBottomDrawerPlugin.self` on Player.swift#208

3 - Interact with the red view by tap and drag gesture

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -113,6 +113,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         containers.forEach(renderContainer)
         addToContainer()
         parentView?.bringSubviewToFront(overlayView)
+        overlayView.clipsToBounds = true
     }
 
     #if os(tvOS)
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -109,7 +109,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
 
     open override func render() {
-        parentView?.addSubviewMatchingConstraints(overlayView)
+        view.addSubviewMatchingConstraints(overlayView)
         containers.forEach(renderContainer)
         addToContainer()
         parentView?.bringSubviewToFront(overlayView)
@@ -140,6 +140,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         plugins
             .compactMap { $0 as? OverlayPlugin }
             .forEach(render)
+        view.bringSubviewToFront(overlayView)
     }
     #endif
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -125,6 +125,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     private func renderCorePlugins() {
         plugins
             .filter { $0.isNotMediaControlElement }
+            .filter { $0.isNotOverlayPlugin}
             .forEach(render)
     }
 
@@ -254,6 +255,10 @@ fileprivate extension Plugin {
     #if os(iOS)
     var isNotMediaControlElement: Bool {
         return !(self is MediaControl.Element)
+    }
+
+    var isNotOverlayPlugin: Bool {
+        return !(self is OverlayPlugin)
     }
     #endif
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -112,7 +112,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         view.addSubviewMatchingConstraints(overlayView)
         containers.forEach(renderContainer)
         addToContainer()
-        parentView?.bringSubviewToFront(overlayView)
+        view.bringSubviewToFront(overlayView)
         overlayView.clipsToBounds = true
     }
 
--- a/Sources/Clappr/Classes/Enum/Event.swift
+++ b/Sources/Clappr/Classes/Enum/Event.swift
@@ -67,5 +67,4 @@ public enum Event: String, CaseIterable {
     case hideDrawerPlugin = "Clappr:hideDrawerPlugin"
     case willHideDrawerPlugin = "Clappr:willHideDrawerPlugin"
     case didHideDrawerPlugin = "Clappr:didHideDrawerPlugin"
-    case didDragDrawer = "Clappr:didDragDrawer"
 }
--- a/Sources/Clappr/Classes/Enum/Event.swift
+++ b/Sources/Clappr/Classes/Enum/Event.swift
@@ -67,4 +67,5 @@ public enum Event: String, CaseIterable {
     case hideDrawerPlugin = "Clappr:hideDrawerPlugin"
     case willHideDrawerPlugin = "Clappr:willHideDrawerPlugin"
     case didHideDrawerPlugin = "Clappr:didHideDrawerPlugin"
+    case didDragDrawer = "Clappr:didDragDrawer"
 }
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
@@ -1,7 +1,8 @@
 class BottomDrawerPlugin: DrawerPlugin {
+    required init(context: UIObject) {
+        super.init(context: context)
 
-    open class override var name: String {
-        return "BottomDrawerPlugin"
+        self.addTapGesture()
     }
 
     override var position: DrawerPlugin.Position {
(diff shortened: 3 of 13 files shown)


Sources/Clappr/Classes/Enum/Event.swift
```diff
@@ -67,4 +67,5 @@ public enum Event: String, CaseIterable {
     case hideDrawerPlugin = "Clappr:hideDrawerPlugin"
     case willHideDrawerPlugin = "Clappr:willHideDrawerPlugin"
     case didHideDrawerPlugin = "Clappr:didHideDrawerPlugin"
+    case didDragDrawer = "Clappr:didDragDrawer"
```
caiobzen: Do we want to make it an internal event? I'm not sure if the apps will want to listen to this event 🤷‍♂

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -0,0 +1,126 @@
...
+        showDrawerPlugin()
+        hideMediaControl()
+    }
+
+    private func showDrawerPlugin() {
+        core?.trigger(.showDrawerPlugin)
+    }
+
+    private func hideMediaControl() {
+        activeContainer?.trigger(.disableMediaControl)
```
caiobzen: since MediaControl listens to `. didShowDrawerPlugin ` in order to hide himself:
```swift
listenTo(core, event: .didShowDrawerPlugin) { [weak self] _ in
    self?.hide()
}
```

I think we can test out if we really need to trigger the `disableMediaControl ` event

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private func hideDrawerPlugin() {
+        core?.trigger(.hideDrawerPlugin)
+        activeContainer?.trigger(.enableMediaControl)
```
caiobzen: Same here! `MediaControl` can listen to `.hideDrawerPlugin` in order to show himself again

Sources/Clappr/Classes/Enum/Event.swift
paulogama14: Done.

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
paulogama14: Nice.

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
paulogama14: Great. 
=]

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        gesture.setTranslation(.zero, in: view)
+    }
+
+    private func handleGestureChange(with gesture: UIPanGestureRecognizer) {
+        guard let view = gesture.view else { return }
+
+        if canDrag(with: gesture.newYCoordinate) {
+            let portionShown = initialCenterY - gesture.newYCoordinate
+            let alpha = hiddenHeight / portionShown * 0.08
```
BruunoMM: Hey, @paulogama14.
Is this variable to indicate the media control alpha ? If so, do you think we could call it something like `mediaControlAlpha` ?
It took me a bit to understand what it was for.

paulogama14: Yeah, I think we can change to another name, more expressive. But, with `mediaControlAlpha` we couple the BottomDrawerPlugin with MediaControl even indirectly. What do you think?

BruunoMM: You mean that if there's something else that we want to hide, then the parameter would not represent that either?
I couldn't think about a specific name, but I think that this is the alpha of the elements we want to hide, right? Any ideas?

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -109,10 +109,11 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
 
     open override func render() {
-        parentView?.addSubviewMatchingConstraints(overlayView)
+        view.addSubviewMatchingConstraints(overlayView)
         containers.forEach(renderContainer)
         addToContainer()
-        parentView?.bringSubviewToFront(overlayView)
+        view.bringSubviewToFront(overlayView)
+        overlayView.clipsToBounds = true
```
BruunoMM: Just a question, why do we need to set this property?

caiobzen: Clips to bounds is what makes the drawer disappear when it goes of of the screen. Clips do bounds doesn’t let subviews to be presented outside the parent view

caiobzen: Clips to bounds is what makes the drawer disappear when it goes of of the screen. Clips do bounds doesn’t let subviews to be presented outside the parent view

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
paulogama14: I've changed to `mediaControlAlpha`. 
=]

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
```diff
@@ -32,36 +38,53 @@ open class DrawerPlugin: OverlayPlugin {
...
+        bindMediaControlEvents(context: core)
+        bindDrawerEvents(context: core)
     }
 
-    private func bindMediaControlEvents() {
-        guard let core = core else { return }
+    private func bindCoreEvents(context: UIObject) {
+        let eventsToRender: [Event] = [.didEnterFullscreen, .didExitFullscreen, .didChangeScreenOrientation]
+
+        listenTo(context, eventName: InternalEvent.didTappedCore.rawValue) { [weak self] _ in
```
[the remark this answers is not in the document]
paulogama14: Nice.
Done
\o

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -51,6 +51,19 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
...
+            guard let alpha = info?["alpha"] as? CGFloat else { return }
+            self?.view.alpha = alpha
+        }
+
+        listenTo(core, event: .didShowDrawerPlugin) { [weak self] _ in
+            self?.hide()
+        }
+
+        listenTo(core, event: .didHideDrawerPlugin) { [weak self] _ in
+            self?.show()
```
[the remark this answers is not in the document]
paulogama14: This thing was solved on `didComplete` and doesn't need to change.

Sources/Clappr/Classes/Base/Core.swift
paulogama14: =]

Tests/Clappr_Tests/Classes/Plugin/Core/Drawer/BottomDrawerPluginTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                }
+
+                it("has a view frame size equals to his size") {
+                    plugin.render()
+
+                    expect(plugin.view.frame.size).to(equal(CGSize(width: 320, height: 50)))
+                }
+
+                context("when plugin request a height greater then the limit") {
+                    it("uses the height limit insted") {
```
BruunoMM: There are two typos in this test description:
In the context description, it should be ```context("when plugin requests a height greater than the limit") ```
```suggestion
                    it("uses the height limit instead") {
```

Tests/Clappr_Tests/Classes/Plugin/Core/Drawer/BottomDrawerPluginTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                context("when the hideDrawer event is triggered") {
+                    it("push down the drawer to core height") {
+                        core.trigger(.hideDrawerPlugin)
+
+                        expect(plugin.view.frame.origin).to(equal(CGPoint(x: .zero, y: 100)))
+                    }
+                }
+
+                context("when the didTappedCore event is triggered") {
+                    it("calls hideDrawer event") {
```
BruunoMM: I think this should be tested in the `DrawerPluginTests`, since it's an action performed by the `DrawerPlugin` class.
What do you think?

PR #351, opened 2019-10-11
AVFoundationPlayback code climate issues fix

caiobzen:
## 🏁  Goal: 
To solve some of the minor Code Climate issues we have in this file.
We know that AVFoundationPlayback needs a deeper refactoring strategy, so this pull request is aiming to help us out in the future and not to solve the root of the problem.

## ✅  How to test:
- Since this is just some refactoring, run the tests and nothing should break.
- You can even try out some backward/forward seek since there were some changes around this code.

Captura de Tela 2019-10-11 às 16 54 20

### DIFF
--- a/Sources/Clappr/Classes/Extension/AVPlayerItem+Ext.swift
+++ b/Sources/Clappr/Classes/Extension/AVPlayerItem+Ext.swift
@@ -4,4 +4,14 @@ extension AVPlayerItem {
     func isFinished(with threshold: TimeInterval = 2.0) -> Bool {
         return fabs(CMTimeGetSeconds(duration) - CMTimeGetSeconds(currentTime())) <= threshold
     }
+
+    func seek(to timeInterval: TimeInterval, _ completion: (() -> Void)? = nil) {
+        seek(
+            to: timeInterval.seek().time,
+            toleranceBefore: timeInterval.seek().tolerance,
+            toleranceAfter: timeInterval.seek().tolerance
+        ) { success in
+            if success { completion?() }
+        }
+    }
 }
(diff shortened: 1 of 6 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -203,19 +203,15 @@ open class AVFoundationPlayback: Playback {
 
     open override func play() {
         guard canPlay else { return }
-
-        if player == nil {
-            setupPlayer()
-        }
-
+        setupPlayerIfNeeded()
```
paulogama14: It's a good point in a better way, but I think we are adding more than one responsibility to this method.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -203,19 +203,15 @@ open class AVFoundationPlayback: Playback {
...
         trigger(.willPlay)
         player?.play()
+        updateInitialStateIfNeeded()
```
paulogama14: Some as above.

Sources/Clappr/Classes/Extension/AVPlayerItem+Ext.swift
```diff
@@ -4,4 +4,14 @@ extension AVPlayerItem {
     func isFinished(with threshold: TimeInterval = 2.0) -> Bool {
         return fabs(CMTimeGetSeconds(duration) - CMTimeGetSeconds(currentTime())) <= threshold
     }
+
+    func seek(to timeInterval: TimeInterval, _ completion: (() -> Void)? = nil) {
```
paulogama14: Is it valid to test this separately?

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
caiobzen: the more than one responsibility was already there. The goal of this pull request is to reduce the `cognitive complexity` around it.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
caiobzen: Same as above

PR #353, opened 2019-10-16
Fix events of click in subview of drawer plugin

johnmartinstw:
## 🏁  Goal: 
When we have subviews in a plugin of type BottomDrawerPlugin the events of click dont work because of the gestures recognizers. We just set the `cancelsTouchesInView` to false to make this work. And make minor changes in `MediaControl` removing the flag `showOnDrawerHide`and validating if the media control can show when drawer hides by the `PlaybackState`.

## ✅  How to test:
1 - Add this code on BottomDrawerPlugin.swift#127 

```swift
    }

        return 32.0
    }

    override func render() {
        super.render()
        let button = UIButton()
        button.backgroundColor = .gray
        button.setTitle("Click me", for: .normal)
        button.addTarget(self, action: #selector(cliked), for: .touchUpInside)
        view.addSubviewMatchingConstraints(button)
    }

     @objc func cliked() {
          print("Yes, clicked")
    }
}
```
2 - Add `CustomBottomDrawerPlugin.self` on Player.swift#208

3 - Interact with the view by tap and drag gesture.
4 - Interact with the button by clicking and see the print.
5 - Run the test suite

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
@@ -28,8 +28,8 @@ open class BottomDrawerPlugin: DrawerPlugin {
     required public init(context: UIObject) {
         super.init(context: context)
 
-        addTapGesture()
-        addDragGesture()
+        addGesture(UITapGestureRecognizer(target: self, action: #selector(didTapView)))
+        addGesture(UIPanGestureRecognizer(target: self, action: #selector(onDragView)))
     }
 
     override open func render() {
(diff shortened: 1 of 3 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -128,12 +125,13 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
         }
 
         listenTo(context, event: .didHideDrawerPlugin) { [weak self] _ in
-            guard self?.showOnDrawerHide == true else { return }
+            let statesToShow: [PlaybackState] = [.playing, .paused]
+
+            guard let state = self?.activePlayback?.state, statesToShow.contains(state) else { return }
```
BruunoMM: Nice! 👍

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -65,13 +65,15 @@ open class BottomDrawerPlugin: DrawerPlugin {
     }
 
     private func addTapGesture() {
-        let gesture = UITapGestureRecognizer(target: self, action: #selector(didTapView))
-        view.addGestureRecognizer(gesture)
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(didTapView))
+        tapGesture.cancelsTouchesInView = false
+        view.addGestureRecognizer(tapGesture)
```
BruunoMM: CodeClimate is complaining about similarity between this method and the one below.
Do you think creating another method would solve this? Something like:
```swift 
    private func addGesture(_ gesture: UIGestureRecognizer) {
        gesture.cancelsTouchesInView = false
        view.addGestureRecognizer(gesture)
    }
```
and passing the gesture to it, like:

```swift
    addGesture(UITapGestureRecognizer(target: self, action: #selector(didTapView)))
```

Issue #319, opened 2019-06-24
didshowMediaControlEvent does not trigged


fukemy:
Hi. I trying to find problem, debug in MediaControl.swift, i got this line
```
self?.core?.trigger(Event.didShowMediaControl.rawValue)
```
it triggered, but not called to my delegate:
```
historyPlayer.on(Event.didShowMediaControl) { _ in
            print("on didShowMediaControl")
            self.slidingUpManager.slideEnabled = false
        }
```

Can you help?

PR #354, opened 2019-10-19
Propagates media control events on player

BruunoMM:
This PR fixes issue #319 
## 🏁  Goal: 
To propagate media control events on the player, so that external plugins and such can listen to the `willShowMediaControl`, `didShowMediaControl`, `willHideMediaControl` and `didHideMediaControl` events.

## ✅  How to test:
1 - Interact with the player and check if media control events are being printed on the console as they're triggered.
2 - Run tests.

### DIFF
--- a/Sources/Clappr_iOS/Classes/Base/Player.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Player.swift
@@ -101,7 +101,7 @@ open class Player: BaseObject {
     }
     
     private func bindMediaControlEvents() {
-        let eventsToListen: [Event] = [.willShowMediaControl, .didShowMediaControl, .willHideMediaControl, .didHideMediaControl]
+        let eventsToListen = [Event.willShowMediaControl, .didShowMediaControl, .willHideMediaControl, .didHideMediaControl]
         
         eventsToListen.forEach { event in
             core?.on(event.rawValue) { [weak self] _ in self?.trigger(event.rawValue) }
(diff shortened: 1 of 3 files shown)


Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -97,6 +97,15 @@ open class Player: BaseObject {
         core?.on(Event.didChangeActivePlayback.rawValue) { [weak self] _ in self?.bindPlaybackEvents() }
         core?.on(InternalEvent.userRequestEnterInFullscreen.rawValue) { [weak self] (info: EventUserInfo) in self?.trigger(Event.requestFullscreen.rawValue, userInfo: info) }
         core?.on(InternalEvent.userRequestExitFullscreen.rawValue) { [weak self] (info: EventUserInfo) in self?.trigger(Event.exitFullscreen.rawValue, userInfo: info) }
+        bindMediaControlEvents()
```
caiobzen: I'd rather move it between line `84` and line `85`, because it looks like `bindMediaControlEvents` is now a side effect of calling `bindCoreEvents`. We know that both are core dependant, but semantically, they are different "layers".

paulogama14: I think we can put this method call `player.init`, like the other binds. What do you think?

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -97,6 +97,15 @@ open class Player: BaseObject {
...
+    }
+    
+    private func bindMediaControlEvents() {
+        let eventsToListen: [Event] = [.willShowMediaControl, .didShowMediaControl, .willHideMediaControl, .didHideMediaControl]
```
rcavaltw: No strong opinion, but this has a few less chars
```suggestion
        let eventsToListen = [Event.willShowMediaControl, .didShowMediaControl, .willHideMediaControl, .didHideMediaControl]
```

Sources/Clappr_iOS/Classes/Base/Player.swift
BruunoMM: Yeah, it makes perfect sense.

PR #356, opened 2019-10-25
Release Clappr from CI

BruunoMM:
## 🏁  Goal: 
It makes possible to make a release directly from the CI machine, deducting the new version number from the project's `Info.plist` file.

## ✅  How to test:
We would need to generate a release in order to test this lane.


lfonseca:
One of the initials concerns about this feature was to assure that the release should be done using the the revision that originated the pipeline. Was it considered?

paulogama14:
> One of the initial concerns about this feature was to assure that the release should be done using the revision that originated the pipeline. Was it considered?

Nice, we've added a verify before the script execution.

PR #357, opened 2019-10-31
Prevent quick-seek on overlay plugins

caiobzen:
## 🏁 Goal:
- To prevent quick seek to overpass the overlay plugins while presenting them

## ✅ How to test:
1. Create a dummy DrawerPlugin:

```swift
    }

        return 32.0
    }

    override func render() {
        super.render()

        view.backgroundColor = .red
    }
}
```
and add it to the list of built in plugins `(Player.swift file)`

2. Run the sample app
3. Tap on the red area to present the drawer plugin
4. Try to double tap the screen like if you were to quick seek forward or backward
5. Nothing should happen

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
@@ -29,6 +29,7 @@ open class DrawerPlugin: OverlayPlugin {
         didSet {
             let event: Event = isClosed ? .didHideDrawerPlugin : .didShowDrawerPlugin
             core?.trigger(event)
+            isClosed ? hide() : show()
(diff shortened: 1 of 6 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/DrawerPlugin.swift
```diff
@@ -29,6 +29,7 @@ open class DrawerPlugin: OverlayPlugin {
         didSet {
             let event: Event = isClosed ? .didHideDrawerPlugin : .didShowDrawerPlugin
             core?.trigger(event)
+            isClosed ? hide() : show()
```
paulogama14: I think that we can group this logic, once the rule is verify if plugind isClosed. What do you think?

PR #358, opened 2019-11-11
Add the current live date property to playback

caiobzen:
## 🏁 Goal:
As we need to know what time is the current live date, we are adding a property that is meant to expose the current live time of a given video.

## ✅ How to test:
1. Since this property is not used in the current player sample app, please run the tests and see that nothing breaks.

### DIFF
--- a/Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackTests.swift
+++ b/Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackTests.swift
@@ -489,6 +489,22 @@ class AVFoundationPlaybackTests: QuickSpec {
                         expect(playback.currentDate).to(equal(date))
                     }
                 }
+
+                describe("#currentLiveDate") {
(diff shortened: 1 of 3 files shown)


Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackTests.swift
```diff
@@ -489,6 +489,46 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                        it("returns the currentLiveDate of the video") {
+                            let playback = StubbedLiveDatePlayback(options: [:])
+                            playback.player = player
+                            let currentDate = Date()
+
+                            item.set(currentDate: currentDate)
+                            playback._position = 30
+                            playback._duration = 1000
+
+                            expect(currentDate.timeIntervalSince1970).to(beLessThan(playback.currentLiveDate?.timeIntervalSince1970))
```
lfonseca: Nice!

PR #362, opened 2019-12-11
Trigger events to identify states

BruunoMM:
# Goal
To trigger state events right after some user action, such as seek, play and pause.

# How To Test
1. Run the sample app

#### Scenario 1 | Paused
1. Pause the video
1. Seek to any position
1. Check if there was a `on Pause` in the console right after the `on didSeek`.

#### Scenario 2 | Playing
1. Play the video
1. Seek to any position
1. Check if there was a `on Play` in the console right after the `on didSeek`.

#### Scenario 3 | Stalling
1. Force the player to buffer (This can be done by simulating a poor connection)
1. Seek to any position
1. Check if there was a `on Stalling` in the console right after the `on didSeek`.

*We've commented a test because the fix to make it succeed will come in another PR.*

### DIFF
--- a/Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackStateMachineEventsTests.swift
+++ b/Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackStateMachineEventsTests.swift
@@ -73,33 +73,33 @@ class AVFoundationPlaybackStateMachineEventsTests: QuickSpec {
                     }
                 }
 
-//                context("when play and seek to end") {
-//                    it("triggers events following the state machine pattern") {
-//                        let options = [kSourceUrl: "http://clappr.sample/master.m3u8"]
-//                        let playback = AVFoundationPlayback(options: options)
-//                        let expectedEvents: [Event] = [
-//                            .ready, .willPlay, .stalling, .willPlay, .stalling, .playing,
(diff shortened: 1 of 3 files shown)


Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackStateMachineEventsTests.swift
```diff
@@ -72,14 +73,41 @@ class AVFoundationPlaybackStateMachineEventsTests: QuickSpec {
                     }
                 }
 
-                context("when play and seek to end") {
+//                context("when play and seek to end") {
```
johnmartinstw: Will we leave this block of code commented? We want this?

dezouzart: we made a few tests with the stalling event adjustment (like bruno explained last tech session). And since we slitted the PR, we had to comment "correct tests" which are failing. We'll uncomment those when the other PR comes.
But if you want to, we can just delete it and then rewrite it on stalling event adjustment PR. Which do you prefer?

johnmartinstw: I did not know that. We can leave it like this then. ;)

PR #365, opened 2020-01-09
Fix QuickSeek when media control is visible

BruunoMM:
Goal
-
To make `QuickSeek` work when the media control is visible and there are `Overlay` plugins.

How To Test
-
1. Play a video
1. Make the media control visible
1. Double tap on the sides
1. The quick seek should take effect

### DIFF
--- a/Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/QuickSeekMediaControlPluginTests.swift
+++ b/Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/QuickSeekMediaControlPluginTests.swift
@@ -77,11 +77,14 @@ class QuickSeekMediaControlPluginTests: QuickSpec {
                         mediaControl.render()
                         playButton.view.layoutIfNeeded()
                         mediaControl.view.layoutIfNeeded()
+                        
                     }
                     
                     context("and it collides with that plugin") {
                         it("does not seek") {
-                            let shouldSeek = quickSeekPlugin.shouldSeek(point: CGPoint(x: 100, y: 100))
+                            let playButtonCenterInMediaControlCoordinate = playButton.view.convert(playButton.view.center, to: mediaControl.mediaControlView)
+                            
+                            let shouldSeek = quickSeekPlugin.shouldSeek(point: playButtonCenterInMediaControlCoordinate)
                             
                             expect(shouldSeek).to(beFalse())
                         }
@@ -89,7 +92,10 @@ class QuickSeekMediaControlPluginTests: QuickSpec {
                     
                     context("and it does not collide with that plugin") {
                         it("does seek") {
-                            let shouldSeek = quickSeekPlugin.shouldSeek(point: CGPoint(x: 260, y: 100))
+                            let pointOutsidePlayButton = CGPoint(x: playButton.view.frame.width + 1, y: playButton.view.frame.height + 1)
+                            let outsidePointInMediaControlCoordinate = playButton.view.convert(pointOutsidePlayButton, to: mediaControl.mediaControlView)
+                            
+                            let shouldSeek = quickSeekPlugin.shouldSeek(point: outsidePointInMediaControlCoordinate)
                             
                             expect(shouldSeek).to(beTrue())
                         }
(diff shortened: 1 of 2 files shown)


Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/QuickSeekMediaControlPluginTests.swift
```diff
@@ -79,6 +83,30 @@ class QuickSeekMediaControlPluginTests: QuickSpec {
                     }
                 }
                 
+                context("and there are overlay plugins") {
+                    it("ignores them and seeks") {
+                        mediaControl.view.layoutIfNeeded()
+                        overlayPlugin.view.layoutIfNeeded()
+                        playButton.view.layoutIfNeeded()
+                        
+                        let shouldSeek = quickSeekPlugin.shouldSeek(point: CGPoint(x: 260, y: 100))
```
torives: @BruunoMM, I was wondering if it would make sense to name this point (and other points created in this file) using a constant or, if the point belongs to a specific plugin, use a syntax like `plugin.view.center.x`, for example. I think that naming it would provide more context and therefore readability to the test, but I'm not sure if this is relevant enough, though. 

What do you think?

BruunoMM: That is a cool suggestion. It would really improve the readability of the test.
Yes, can do! o/

pauloSantosMobileTW:
When a video ends and start again on steve, if I open config modal I can quick seek. Can you check about that?

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/QuickSeekMediaControlPluginTests.swift
```diff
@@ -68,14 +67,64 @@ class QuickSeekMediaControlPluginTests: QuickSpec {
...
+                            
+                            let shouldSeek = quickSeekPlugin.shouldSeek(point: playButtonCenterInMediaControlCoordinate)
+                            
+                            expect(shouldSeek).to(beTrue())
+                        }
+                    }
+                }
+                
+                context("and there are overlay plugins") {
+                    it("ignores them and seeks") {
```
pauloSantosMobileTW: This phrase makes sense? If we have an overlay plugin like the config screen it should seek?

BruunoMM: Actually it should not seek, but we do have another way of handling that.
QuickSeekPlugin listens to the `didShowOverlayPlugin` and `didHideOverlayPlugin` events, and when they're triggered it removes and adds the gesture, respectively.

BruunoMM: We have a different way of handling `Overlay Plugins`, by listening to some show and hide events triggered by them, so it should not be necessary to take that in consideration here. We only added that test because we must ignore those kind of plugins in the `filteredOutPlugins` method, knowing that regardless of being visible or not, they still exist in the plugins list.

BruunoMM: Maybe an example can make things clearer:
- If there is a visible OverlayPlugin, then we won't seek because we won't have a gesture recognizer to capture the taps. If there isn't a visible one, then we should ignore them in the filtering and seek.

BruunoMM:
> When a video ends and start again on steve, if I open config modal I can quick seek. Can you check about that?

This issue will be addressed in another PR. Thanks for the heads up!

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/QuickSeekMediaControlPluginTests.swift
pauloSantosMobileTW: The context look like the opposite to me:
- If there is a visible OverlayPlugin then we should ignore it in the filtering and seek.

Make sense?

PR #366, opened 2020-01-20
Correct the broken unit tests

johnmartinstw:
## Goal
Correct the broken tests on `ContainerTests` and  `BottomDrawerPluginTests`
## How to test
1. Run test suit
2. Check if all tests passed

### DIFF
--- a/Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
+++ b/Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
@@ -10,6 +10,11 @@ class ContainerTests: QuickSpec {
             var container: Container!
 
             describe("#Init") {
+
+                beforeEach {
+                    Loader.shared.resetPlaybacks()
+                }
+
                 context("with a invalid resource") {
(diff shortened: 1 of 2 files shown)


Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
```diff
@@ -10,6 +10,11 @@ class ContainerTests: QuickSpec {
             var container: Container!
 
             describe("#Init") {
+
+                beforeEach {
+                    Loader.shared.resetPlaybacks()
```
lfonseca: As far as I can see the bigger problem in these tests was this missing logic. Maybe we can discuss in our next tech session about a better approach to avoid this mistake in the future.

lfonseca:
Great job, guys! Good to go to me.

Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
johnmartinstw: Yes! I agree completely!

PR #367, opened 2020-01-29
Remove memory leaks

BruunoMM:
Goal
---
Remove memory leaks to avoid the app being slow or crashing.

How to Test
---
1. Run the Clappr Example
1. Proceed to the PlayerViewController
1. Tap the `Recreate Player` button multiple times
1. Click on the `Debug Memory Graph` button in the bottom Debug Area of Xcode.
Screen Shot 2020-01-30 at 11 02 42

1. Click the `Show only content from workspace` button to filter the list.
Screen Shot 2020-01-30 at 11 05 52

1. Check if there's any class from Clappr in the memory leak list.

### DIFF
--- a/Example/Clappr/ViewController.swift
+++ b/Example/Clappr/ViewController.swift
@@ -14,11 +14,7 @@ class ViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        player = Player(options: options)
-
-        listenToPlayerEvents()
-
-        player.attachTo(playerContainer, controller: self)
+        createPlayer()
     }
 
     @objc func listenToPlayerEvents() {
@@ -48,14 +44,14 @@ class ViewController: UIViewController {
         
         player.on(Event.didHideMediaControl) { _ in print("on didHideMediaControl") }
 
-        player.on(Event.requestFullscreen) { _ in
+        player.on(Event.requestFullscreen) { [weak self] _ in
             print("on requestFullscreen")
-            self.onRequestFullscreen()
+            self?.onRequestFullscreen()
         }
 
-        player.on(Event.exitFullscreen) { _ in
+        player.on(Event.exitFullscreen) { [weak self] _ in
             print("on exitFullscreen")
-            self.onExitFullscreen()
+            self?.onExitFullscreen()
         }
     }
 
@@ -92,5 +88,20 @@ class ViewController: UIViewController {
         let alertViewController = UIAlertController(title: title, message: message, preferredStyle: .alert)
         alertViewController.addAction(UIAlertAction(title: "ok", style: UIAlertAction.Style.default, handler: nil))
         self.navigationController?.present(alertViewController, animated: true, completion: nil)
+    }
+    
+    private func createPlayer() {
+        player = Player(options: options)
+
+        listenToPlayerEvents()
+
+        player.attachTo(playerContainer, controller: self)
+        player.play()
+    }
+    
+    @IBAction func recreatePlayer(_ sender: Any) {
+        player.destroy()
+        player = nil
+        createPlayer()
     }
 }
--- a/Example/Clappr/ViewController.swift
+++ b/Example/Clappr/ViewController.swift
@@ -96,12 +96,12 @@ class ViewController: UIViewController {
         listenToPlayerEvents()
 
         player.attachTo(playerContainer, controller: self)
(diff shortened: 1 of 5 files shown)


Example/Clappr/ViewController.swift
```diff
@@ -93,4 +89,19 @@ class ViewController: UIViewController {
...
         self.navigationController?.present(alertViewController, animated: true, completion: nil)
     }
+    
+    private func createPlayer() {
+        player = Player(options: options)
+
+        listenToPlayerEvents()
+
+        player.attachTo(playerContainer, controller: self)
+        player.play()
```
paulogama14: I think we can avoid calling a play because we are changing the behavior of the player. With this new state, the player has autoplay and we lost the Poster in the beginning. Does it make sense?

Sources/Clappr_iOS/Classes/Animation/QuickSeekAnimation.swift
```diff
@@ -1,5 +1,5 @@
 class QuickSeekAnimation {
-    private var core: Core?
+    private weak var core: Core?
```
paulogama14: I'm analyzing the code and found the use of Core is to provide a view that receives animation from the `QuickSeekPlugin`. Is there another way to avoid this bad behavior (retain cycle on this object)? 
One suggestion maybe is split the responsibility between Plugin and the View, and the View do not know the Core. What do you think?

Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
```diff
@@ -3,7 +3,7 @@ class SeekBubble: UIView {
     private var images: [UIImageView] = []
     private var text: String!
     
-    private var parentView: UIView!
+    private weak var parentView: UIView!
```
paulogama14: Same idea as above.

Example/Clappr/ViewController.swift
BruunoMM: Yes! This was just to facilitate the testing. It can be undone.

Sources/Clappr_iOS/Classes/Animation/QuickSeekAnimation.swift
BruunoMM: Yes. We actually don't need to store the reference to the core nor the view here, we can just pass it through.

Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
BruunoMM: Actually this one is already a view and it needs the reference to the parentView, since we're doing things this way with this specific plugin.
The thing that causes the retain cylce here is that in the view hierarchy, the superview has a reference to this view and vice-versa, that's why the reference must be a weak one.

If you have suggestions on how to overcome this by other means, please let me know.

Sources/Clappr_iOS/Classes/Animation/QuickSeekAnimation.swift
helbertgs: I'm thinking about the same approach. Make more sense

BruunoMM: It's done! o/

Sources/Clappr_iOS/Classes/Animation/QuickSeekAnimation.swift
```diff
@@ -1,13 +1,11 @@
...
     private var seekLeftBubble = SeekBubble()
     private var seekRightBubble = SeekBubble()
     
-    init(_ core: Core?) {
-        self.core = core
-        if let coreView = core?.view {
-            seekLeftBubble.setup(within: coreView, bubbleSide: .left)
-            seekRightBubble.setup(within: coreView, bubbleSide: .right)
+    init(_ view: UIView?) {
+        if let parentView = view {
```
paulogama14: In my opinion, it's a better approach the `guard let` in this case. What do you think?

helbertgs: I think the same.

BruunoMM: Done! :v:

PR #369, opened 2020-03-02
New quality metrics

johnmartinstw:
### Goal
Creating and exposing new quality metrics from `AVFoundationPlayback`. 

### How to test
- Run test suite.


Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
```diff
@@ -0,0 +1,117 @@
...
+                            name: NSNotification.Name.AVPlayerItemNewAccessLogEntry,
+                            object: player.currentItem
+                        )
+
+                        expect(avfoundationPlayback.droppedFrames).toEventually(equal(2))
+                    }
+                }
+            }
+
+            context("when call play") {
```
pauloSantosMobileTW: Make sense move te common code to the context and have only the "expect" inside each "it" ?

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -30,6 +31,15 @@ open class AVFoundationPlayback: Playback {
 
     private var observers = [NSKeyValueObservation]()
 
+    private var lastLogEvent: AVPlayerItemAccessLogEvent? { player?.currentItem?.accessLog()?.events.last }
+
+    open var bitrate: Double? { lastLogEvent?.indicatedBitrate }
+    open var bandwidth: Double? { lastLogEvent?.observedBitrate }
+    open var averageBitrate: Double? { lastLogEvent?.averageVideoBitrate }
+    open var droppedFrames: Int? { lastLogEvent?.numberOfDroppedVideoFrames }
+    open var decodedFrames: Int? { -1 }
```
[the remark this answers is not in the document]
paulogama14: Yes, we made a research about this value on `AVPlayer` and didn't get a safe and reliable way to expose it. There is no documentation on AVPlayer context and the other suggestions are expensives calcules and without confidence. Do you have any idea?

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
paulogama14: I don't like this approach, in my opinion the test should have an explicit behavior. We follow the AAA pattern for tests and omitting code can brake this rule. What do you think?

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                stubsDescriptor?.name = "StubToHighlineVideo.mp4"
+            }
+
+            afterEach {
+                OHTTPStubsHelper.removeStub(with: stubsDescriptor)
+            }
+
+            context("when receives AVPlayerItemNewAccessLogEntry event") {
```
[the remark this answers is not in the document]
paulogama14: What do you mean? The stub for contexts?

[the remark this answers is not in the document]
pauloSantosMobileTW: The assemble code, something like that:
`                      let options = [
                            kSourceUrl: "http://clappr.io/highline.mp4"
                        ]
                        let avfoundationPlayback = AVFoundationPlayback(options: options)
                        let accessLog = AccessLogEventMock()
                        let player = PlayerMock(accessLogEvent: accessLog)
                        avfoundationPlayback.player = player
                        avfoundationPlayback.addObservers()

                        it("changes the bitrate") {
                            accessLog.setIndicatedBitrate(7.0)
                            NotificationCenter.default.post(
                                name: NSNotification.Name.AVPlayerItemNewAccessLogEntry,
                                object: player.currentItem
                        )

                        expect(avfoundationPlayback.bitrate).toEventually(equal(7.0))
`

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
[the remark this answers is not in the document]
pauloSantosMobileTW: LGTM this aproach, I can not find another solution too for that, thanks for sharing.

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
pauloSantosMobileTW: I think the context can be saw as part of the AAA pattern, I always imagine like I am reading a book when reading a test and the duplicity makes the read harder in this case to me.

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
```diff
@@ -0,0 +1,117 @@
+import Quick
+import Nimble
+import OHHTTPStubs
+
+@testable import Clappr
+
+class AVFoundationPlaybackQualityMetricsTests: QuickSpec {
+    override func spec() {
+        describe(".AVFoundationPlaybackQualityMetricsTests") {
```
lfonseca: Are these tests supposed to be integration ones?

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -323,8 +321,8 @@ open class AVFoundationPlayback: Playback {
     @objc func onAccessLogEntry(notification: NSNotification?) {
         guard lastBitrate != bitrate else { return }
         lastBitrate = bitrate
-        if let newBitrate = lastBitrate, !newBitrate.isNaN {
-            trigger(.didUpdateBitrate, userInfo: ["bitrate": newBitrate])
+        if let lastBitrate = lastBitrate, !lastBitrate.isNaN {
```
lfonseca: 👏 👏

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
paulogama14: No, we've changed the header from response and now the stub are working. Could you check, please?

lfonseca: So maybe you should call the `onAccessLogEntry` method directly and config some spy object to catch the event sent.

paulogama14: Nice =]
Now we are checking with a spy object the event sent, the other tests don't have a `kSourceURL`.

PR #370, opened 2020-03-11
Create a unique callback to handle error user info

BruunoMM:
Goal
---
Create a separate callback for the error event.

How to test
---
- Run test suite for both iOS and tvOS targets

### DIFF
--- a/Sources/Clappr_iOS/Classes/Base/Player.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Player.swift
@@ -70,16 +70,16 @@ open class Player: BaseObject {
         Logger.logInfo("loading with \(options)", scope: "Clappr")
 
         self.playbackEventsToListen.append(contentsOf:
-            [Event.ready.rawValue, Event.error.rawValue,
-             Event.playing.rawValue, Event.didComplete.rawValue,
-             Event.didPause.rawValue, Event.stalling.rawValue,
-             Event.didStop.rawValue, Event.didUpdateBuffer.rawValue,
-             Event.willPlay.rawValue, Event.didUpdatePosition.rawValue,
-             Event.willPause.rawValue, Event.willStop.rawValue,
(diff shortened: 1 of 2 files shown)


Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -222,21 +221,29 @@ open class Player: AVPlayerViewController {
 
     private func bindPlaybackEvents() {
         if let playback = core?.activePlayback {
+            var listenId = ""
```
paulogama14: Same as above.

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -222,21 +221,29 @@ open class Player: AVPlayerViewController {
 
     private func bindPlaybackEvents() {
         if let playback = core?.activePlayback {
```
paulogama14: Same as above.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -188,18 +188,26 @@ open class Player: BaseObject {
 
     private func bindPlaybackEvents() {
         if let playback = core?.activePlayback {
```
paulogama14: I know this is not from this PR, but we can change this if let to guard let, once we don't have code outside with a nullable core and we can remove one level of indentation.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -188,18 +188,26 @@ open class Player: BaseObject {
...
+            var listenId = ""
```
paulogama14: I think we don't need to share this variable. We can keep it inside for context. What do you think?

BruunoMM: We share so we don't have to define it every iteration and then again outside the for loop.

PR #372, opened 2020-03-13
Allow the player to be muted on iOS

torives:
### 🏁 Goal
Allow for the iOS Player to be muted. This was already possible on the tvOS target, this PR uses the same strategy on iOS and add some tests.

### ✅ How to test
- Run the test suite


Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -163,6 +163,10 @@ open class Player: BaseObject {
     open func seek(_ timeInterval: TimeInterval) {
         core?.activePlayback?.seek(timeInterval)
     }
+    
+    open func mute(enabled: Bool) {
```
paulogama14: Hey @torives, maybe we can expose two methods and split into unique responsibility. What do you think?

PR #373, opened 2020-03-18
Prevent play when application is not active

pauloSantosMobileTW:
# Goal
Prevent the video playing while the application is in background.

# How to test
- Open the sample
- Start a new video
- Go to home screen of the phone before video start playing
- The audio should not play


pauloSantosMobileTW:
This PR was closed since it does not fix the problem.

PR #375, opened 2020-04-02
Reload Container

paulogama14:
# Goal

- Create reload flow on Container
- Clear objects when stop is call
- Expose function on Playbutton
- Fix spinner visibility when pause and stop events is triggered

# How to test

- Try to call reload function with other configurations, like: 

On `Player.swift#150`

```swift
open func reload(sourceUrl: String) {
    core?.activeContainer?.reload(source: sourceUrl)
}
```

On `ViewController.swift#101`

```swift
@IBAction func recreatePlayer(_ sender: Any) {
    player.destroy()
    player = nil
    createPlayer()
        
    let sourceUrl = "https://www.radiantmediaplayer.com/media/bbb-360p.mp4"
    player.reload(sourceUrl: sourceUrl)
    player.play()
}
```

- The other goals were covered by unit tests. You can check there

### DIFF
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -50,6 +50,8 @@ open class Container: UIObject {
     }
     
     open func play() {
+        guard playback?.canPlay ?? false else { return }
+        
         if shouldReload {
             reload()
         } else {
@@ -58,6 +60,8 @@ open class Container: UIObject {
     }
     
     open func pause() {
+        guard playback?.canPause ?? false else { return }
+        
         playback?.pause()
     }
 
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -65,13 +65,13 @@ open class Container: UIObject {
         }
     }
 
-    @objc open func reload(videoId: String) {
+    @objc open func reload(source: String) {
         let mimeType = options[kMimeType] as? String
 
         options[kDefaultSubtitle] = playback?.selectedSubtitle?.language
         options[kDefaultAudioSource] = playback?.selectedAudioSource?.language
 
-        load(videoId, mimeType: mimeType)
+        load(source, mimeType: mimeType)
     }
     
     open override func render() {
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -50,19 +50,19 @@ open class Container: UIObject {
     }
     
     open func play() {
-        guard playback?.canPlay ?? false else { return }
+        guard let activePlayback = playback, activePlayback.canPlay else { return }
         
         if shouldReload {
             reload()
         } else {
-            playback?.play()
+            activePlayback.play()
         }
     }
     
     open func pause() {
-        guard playback?.canPause ?? false else { return }
+        guard let activePlayback = playback, activePlayback.canPause else { return }
         
-        playback?.pause()
+        activePlayback.pause()
     }
 
     @objc open func load(_ source: String, mimeType: String? = nil) {
--- a/Sources/Clappr/Classes/Base/Container.swift
+++ b/Sources/Clappr/Classes/Base/Container.swift
@@ -3,7 +3,8 @@ import Foundation
 open class Container: UIObject {
     private(set) var plugins: [Plugin] = []
     @objc open var sharedData = SharedData()
-    @objc open var shouldReload = false
+    private var shouldReload = false
+    private var source: String?
     @objc open var options: Options {
         didSet {
             trigger(Event.didUpdateOptions)
@@ -43,9 +44,26 @@ open class Container: UIObject {
         view.backgroundColor = .clear
         view.accessibilityIdentifier = "Container"
     }
+    
+    open func setNeedsReload() {
+        shouldReload = true
+    }
+    
+    open func play() {
+        if shouldReload {
+            reload()
+        } else {
+            playback?.play()
+        }
+    }
+    
+    open func pause() {
+        playback?.pause()
+    }
 
     @objc open func load(_ source: String, mimeType: String? = nil) {
-        shouldReload = false
+        self.source = source
+        
         trigger(Event.willLoadSource.rawValue)
 
         options[kSourceUrl] = source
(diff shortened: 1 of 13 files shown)


Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -3,6 +3,7 @@ import Foundation
 open class Container: UIObject {
     private(set) var plugins: [Plugin] = []
     @objc open var sharedData = SharedData()
+    @objc open var shouldReload = false
```
lfonseca, line 6: Where this variable is being used as control?

paulogama14: This is a way to force a reload on Container. You can use this reading on a `PlayButtonPlugin`, for example. What do you think?

Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -63,6 +65,15 @@ open class Container: UIObject {
         }
     }
 
+    @objc open func reload(source: String) {
+        let mimeType = options[kMimeType] as? String
+
+        options[kDefaultSubtitle] = playback?.selectedSubtitle?.language
```
johnmartinstw, line 71: What do you think about separating this set of default options in another function?

Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -63,6 +65,15 @@ open class Container: UIObject {
         }
     }
 
+    @objc open func reload(source: String) {
```
lfonseca, line 68: Why would a `reload` receive another `source` instead of reloading the same received before?

Sources/Clappr/Classes/Base/Container.swift
lfonseca: There is no reason to expose `Container`'s internal state and break its encapsulation to share data among plugins. The design is wrong here. We should rethink it once the only entity that should have this knowledge is `Container` itself.

Sources/Clappr/Classes/Base/Container.swift
pauloSantosMobileTW: Done

Sources/Clappr/Classes/Base/Container.swift
pauloSantosMobileTW: Fixed

Sources/Clappr/Classes/Base/Container.swift
pauloSantosMobileTW: Fixed

Sources/Clappr/Classes/Base/Container.swift
```diff
@@ -63,6 +83,23 @@ open class Container: UIObject {
         }
     }
 
+    @objc open func reload() {
```
lfonseca, line 86: What is the reason for this method being public and present in our published API?

Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
```diff
@@ -353,21 +353,71 @@ class ContainerTests: QuickSpec {
...
+            describe("#reload") {
+                context("options") {
+                    it("changes source URL") {
+                        Loader.shared.resetPlaybacks()
+                        Loader.shared.register(playbacks: [StubPlayback.self])
+                        let container = Container()
+                        container.load(Source.invalid)
+                        
+                        container.setNeedsReload()
+                        container.reload()
```
lfonseca, line 366: A `container.play()` would reflect better the behavior here. What do you think?

Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
```diff
@@ -353,21 +353,71 @@ class ContainerTests: QuickSpec {
                     expect(didResizeValue?.height).toEventually(equal(10))
                 }
             }
+            
+            describe("#reload") {
+                context("options") {
+                    it("changes source URL") {
```
lfonseca, line 359: This description seems to be incorrect.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -64,27 +80,40 @@ open class PlayButton: MediaControl.Element {
...
-            return
-        }
-
-        if playback.state == .paused {
-            button.setImage(playIcon, for: .normal)
-        } else if playback.state == .playing {
-            button.setImage(pauseIcon, for: .normal)
-        }
+    private func pause() {
+        guard let playback = activePlayback, playback.canPause else { return }
```
lfonseca, line 97: Once we are moving `play` and `pause` logic to `Container` in order to improve it's encapsulating it would be interesting to evaluate if this statement should be also there.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -64,27 +80,40 @@ open class PlayButton: MediaControl.Element {
...
+        
+        activeContainer?.pause()
+    }
+    
+    private func play() {
+        guard let playback = activePlayback, playback.canPlay else { return }
```
lfonseca, line 103: Same here.

Sources/Clappr/Classes/Base/Container.swift
pauloSantosMobileTW: Fixed

Tests/Clappr_Tests/Classes/Plugin/Container/ContainerTests.swift
pauloSantosMobileTW: Done

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/PlayButtonTests.swift
```diff
@@ -194,6 +193,140 @@ class PlayButtonTests: QuickSpec {
...
+                    core = CoreStub()
+                    playButton = PlayButton(context: core)
+                    playButton.render()
+                }
+
+                context("when playback state is paused") {
+                    context("can play is true") {
+                        it("calls play") {
+                            core.playbackMock?.state = .paused
+                            core.playbackMock?._canPlay = true
```
lfonseca, line 211: Now that we moved this conditional into `Container` there is no reason for this arrange here.

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/PlayButtonTests.swift
```diff
@@ -194,6 +193,140 @@ class PlayButtonTests: QuickSpec {
...
+
+                            playButton.togglePlayPause()
+                            
+                            expect(core.playbackMock?.didCallPlay).to(beTrue())
+                        }
+                    }
+                    
+                    context("can play is false") {
```
lfonseca, line 219: This test doesn't belong to `PlayButton` but to `Container` now.

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/PlayButtonTests.swift
```diff
@@ -194,6 +193,140 @@ class PlayButtonTests: QuickSpec {
...
+                            core.playbackMock?._canPlay = false
+                                
+                            playButton.togglePlayPause()
+                            
+                            expect(core.playbackMock?.didCallPlay).to(beFalse())
+                        }
+                    }
+                }
+                
+                context("when playback state is playing") {
```
lfonseca, line 256: The same as `play` applies here.

PR #380, opened 2020-05-07
Remove reload on Container

paulogama14:
## Goal

- Remove reload on Container
- Remove play / pause on Container
- Clear `timeObserver` instance on `AVFoundationPlayback` when `removeObservers` is called. (Possible retain cycle in some destroy() scenarios)

## How to Test

- Run all test suite
- On sample app, click in the `Recreate Player` button a lot of times, to ensure that player.destroy() is working well and without any retain cycle.


paulogama14:
We are reverting the [PR 375](https://github.com/clappr/clappr-ios/pull/375) on the [PR 381](https://github.com/clappr/clappr-ios/pull/381). So, we are closing this one.

PR #382, opened 2020-05-13
Feature/live start at

pauloSantosMobileTW:
## Goal
Allow DVR Videos to be initiated in a position before the live edge

## How to test
1. Run tests


Clappr.xcodeproj/xcshareddata/xcschemes/Clappr_Tests.xcscheme
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
paulogama14, line 1: Maybe we can avoid adding this file or is there a special reason for this?

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -264,9 +268,19 @@ open class AVFoundationPlayback: Playback {
...
     private func seekToStartAtIfNeeded() {
-        if canStartAt {
-            seek(startAt)
-        }
+        guard canStartAt else { return }
+        
+        seek(startAt)
+    }
+    
+    private func seekToLiveStartTimeIfNeeded() {
```
paulogama14, line 276: I think we are adding more than one responsibility in this method. An indication of that is the `ifNeeded` in the name.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -264,9 +268,19 @@ open class AVFoundationPlayback: Playback {
...
+        guard canLiveStartAt, let liveStartTime = liveStartTime else { return }
+        
+        let positionToSeek = liveStartTime - epochDvrWindowStart
+        
+        if positionToSeek < 0 || positionToSeek > duration { return }
```
paulogama14, line 281: I think this rule can be encapsulated in a method or var to be more expressive. What do you think?

Clappr.xcodeproj/xcshareddata/xcschemes/Clappr_Tests.xcscheme
```diff
@@ -0,0 +1,52 @@
...
+<Scheme
```
helbertgs, line 2: I think you don't need to merge this file

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
helbertgs: I have the same opinion

Clappr.xcodeproj/xcshareddata/xcschemes/Clappr_Tests.xcscheme
dezouzart: this change is about showing test target on xcode menu. IMHO it helps. But I can revert it if you want to.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
pauloSantosMobileTW: Done

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
pauloSantosMobileTW: Done

Clappr.xcodeproj/xcshareddata/xcschemes/Clappr_Tests.xcscheme
pauloSantosMobileTW: See comment above

Clappr.xcodeproj/xcshareddata/xcschemes/Clappr_Tests.xcscheme
paulogama14: Okey-Dokey.
=]

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -227,6 +227,10 @@ open class AVFoundationPlayback: Playback {
     private var canStartAt: Bool {
         return startAt != 0.0 && playbackType == .vod
     }
+    
+    private var canLiveStartTime: Bool {
+        return isEpochInsideDVRWindow(liveStartTime) && isDvrAvailable
```
BruunoMM, line 232: We could short circuit this if we invert the clause's order. Avoiding doing the math if DVR isn't available.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -264,9 +268,17 @@ open class AVFoundationPlayback: Playback {
...
-        if canStartAt {
-            seek(startAt)
-        }
+        guard canStartAt else { return }
+        
+        seek(startAt)
+    }
+    
+    private func seekToLiveStartTime() {
+        guard let liveStartTime = liveStartTime else { return }
```
BruunoMM, line 277: There's no need to do this check here. You already check that in `canLiveStartTime`

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -227,6 +227,10 @@ open class AVFoundationPlayback: Playback {
     private var canStartAt: Bool {
         return startAt != 0.0 && playbackType == .vod
     }
+    
+    private var canLiveStartTime: Bool {
```
BruunoMM, line 231: What do you think about giving a more descriptive name to this variable? I think the word `time` here does not fit. Something around `canStartLiveAt`, `canLiveStartAtTime` or even `canLiveStartAtGivenTime`.

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackTests.swift
```diff
@@ -780,6 +780,127 @@ class AVFoundationPlaybackTests: QuickSpec {
...
+                                    
+                                    playback.on(Event.didUpdatePosition.rawValue) { userInfo in
+                                        didCallUpdatePosition = true
+                                        liveStartTime = userInfo!["position"] as! Double
+                                    }
+                                    
+                                    playback.handleDvrAvailabilityChange()
+                                    
+                                    expect(didCallUpdatePosition).toEventually(beTrue())
+                                    expect(liveStartTime).toEventually(equal(500))
```
BruunoMM, line 820: Hey guys, I found a bit hard to understand why the result would be 500. Do you think is there a more descriptive way to write this test?

My suggestion would be to pass a `currentDate` to `setupLiveDvr`, so the math to calculate the result (expected value) would be contained inside the test itself. WDYT?

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
pauloSantosMobileTW: Swift does not allow make a calculation with a Double? and a Double, this guard statement is required here

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
pauloSantosMobileTW: Done

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
pauloSantosMobileTW: Done

PR #383, opened 2020-05-14
Clear objects on AVFoundationPlayback.destroy()

paulogama14:
# Goal

- Reset properties on `AVFoundationPlayback.destroy()`
- Introduce `AVFoundationPlaybackStopActionTests`
- Create default error on `AVFoundationPlayback.onFailedToPlayToEndTime`
- Minor code styles refactor

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -299,25 +299,36 @@ open class AVFoundationPlayback: Playback {
             self,
             selector: #selector(playbackDidEnd),
             name: .AVPlayerItemDidPlayToEndTime,
-            object: player.currentItem)
+            object: player.currentItem
+        )
         
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(onAccessLogEntry),
             name: .AVPlayerItemNewAccessLogEntry,
-            object: nil)
+            object: nil
+        )
 
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(onFailedToPlayToEndTime),
             name: .AVPlayerItemFailedToPlayToEndTime,
-            object: nil)
+            object: nil
+        )
+    }
+
+    @objc func playbackDidEnd(notification: NSNotification?) {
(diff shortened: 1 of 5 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -299,25 +299,36 @@ open class AVFoundationPlayback: Playback {
...
+        updateState(.idle)
+        droppedFrames = 0
     }
 
     @objc func onFailedToPlayToEndTime(notification: NSNotification?) {
-        if let error = notification?.userInfo?["AVPlayerItemFailedToPlayToEndTimeErrorKey"] as? Error {
-            trigger(.error, userInfo: ["error": error])
-        }
+        let errorKey = "AVPlayerItemFailedToPlayToEndTimeErrorKey"
+        guard let error = notification?.userInfo?[errorKey] as? NSError else { return }
```
geisonfgf, line 329: It is ok do not send any error in case this cast does not work? Maybe would be better to have a default error for this case.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -482,7 +487,11 @@ open class AVFoundationPlayback: Playback {
         playerLayer = nil
         player?.replaceCurrentItem(with: nil)
         player = nil
+    }
+
+    private func clearObjects() {
```
geisonfgf, line 492: clearObjects do not seem to be the real action performed in this method? It seems to reset Playback properties and so a possible better name to it could be resetPlaybackProperties.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
paulogama14: Sure.
=]

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
paulogama14: Done.

PR #387, opened 2020-05-21
Focus update event

helbertgs:
# Goal

Introduce the possibility of a visual entity to request/release focus on `tvOS`

# How to Test

Run tests

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
+++ b/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
@@ -9,6 +9,16 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
         }
     }
     
+    #if os(tvOS)
+    func updateFocus(userInfo: EventUserInfo) {
+        core?.trigger(.requestFocusUpdate, userInfo: userInfo)
+    }
+    
+    func updateFocus() {
+        core?.trigger(.requestFocusUpdate)
+    }
+    #endif
+    
     open func render() {
         NSException(name: NSExceptionName("RenderNotOverriden"), reason: "UICorePlugins should always override the render method").raise()
     }
--- a/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
+++ b/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
@@ -10,11 +10,11 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
     }
     
     #if os(tvOS)
-    func updateFocus(userInfo: EventUserInfo) {
+    open func updateFocus(userInfo: EventUserInfo) {
         core?.trigger(.requestFocusUpdate, userInfo: userInfo)
     }
     
-    func updateFocus() {
+    open func updateFocus() {
         core?.trigger(.requestFocusUpdate)
     }
     #endif
--- a/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
+++ b/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
@@ -10,13 +10,14 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
     }
     
     #if os(tvOS)
-    open func requestFocusToMe() {
+    open func requestFocus() {
         let info: EventUserInfo = ["viewTag": view.tag]
         core?.trigger(.requestFocus, userInfo: info)
     }
     
-    open func updateFocus() {
-        core?.trigger(.updateFocus)
+    open func releaseFocus() {
+        let info: EventUserInfo = ["viewTag": view.tag]
+        core?.trigger(.releaseFocus, userInfo: info)
     }
     #endif
     
--- a/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
+++ b/Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
@@ -10,12 +10,13 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
     }
     
     #if os(tvOS)
-    open func updateFocus(userInfo: EventUserInfo) {
-        core?.trigger(.requestFocusUpdate, userInfo: userInfo)
+    open func requestFocusToMe() {
+        let info: EventUserInfo = ["viewTag": view.tag]
+        core?.trigger(.requestFocus, userInfo: info)
     }
     
     open func updateFocus() {
-        core?.trigger(.requestFocusUpdate)
+        core?.trigger(.updateFocus)
     }
     #endif
     
--- a/Sources/Clappr_tvOS/Classes/Base/Player.swift
+++ b/Sources/Clappr_tvOS/Classes/Base/Player.swift
@@ -41,6 +41,14 @@ open class Player: AVPlayerViewController {
         }
     }
 
+    override open var preferredFocusEnvironments: [UIFocusEnvironment] {
+        if let button = contentOverlayView?.subviews.first?.subviews[1] as? UIButton {
+            return [button]
+        }
+
+        return super.preferredFocusEnvironments
+    }
+
     public var focusEnvironments: [UIView] { contentOverlayView?.subviews.first?.subviews ?? [] }
 
     open override func shouldUpdateFocus(in context: UIFocusUpdateContext) -> Bool { true }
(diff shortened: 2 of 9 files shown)


Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -41,6 +41,15 @@ open class Player: AVPlayerViewController {
         }
     }
 
+    override open var preferredFocusEnvironments: [UIFocusEnvironment] {
+        let subviews = contentOverlayView?.subviews.first?.subviews
+        if let button = subviews?.first(where: { $0 is UIButton }), button.alpha > 0.1 {
```
torives, line 46: I got the gist we're trying to return the SkipButton as the preferred focus environment, but I didn't quite get why there's this `button.alpha > 0.1` check. Could you guys shine some light on this matter, please?

pauloSantosMobileTW: When we hide or show a action button we do it with the alpha value animated, so if the alpha is greater then 0.1 it is being showed. If we fix it in 1.0 it could have any bugs related to animation.

torives: Then I guess we can extract it to some `isVisible:button:` function (or something in that line), just so the code's intent is clear.

torives: I guess what is missing here is to actually hide the button view with `isHidden = true`. A hidden view should not be processed by the Focus Engine, therefore making this check obsolete.

fcapinho, line 46: I consider that relying only on the visual element type (`UIButton`) to choose the focus is not safe (only works on the current UI implementation and plugins).
Can we pass a UI element identification on the new event (`requestFocusUpdate`) ?

pauloSantosMobileTW: We tried an approach with hidden and it has some inconsistencies, we need to work with alpha here because of the animations of show/hide, if we do not do this the player do not know when to come back the focus to the seek bar when the button is hiding.

pauloSantosMobileTW: We changed it to a generic UIView and receives the UIView reference on the event userInfo, can you check again?

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
```diff
@@ -9,6 +9,16 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
         }
     }
     
+    #if os(tvOS)
+    func updateFocus(userInfo: EventUserInfo) {
```
fcapinho, line 13: Is this functionality applicable to all types of `UIView`s / `UICorePlugin`s ?
Current usage looks like a specific kind of `Plugin` only functionality.

About the name, the intent is not to actually update the focus but rather to request focus for itself.

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -8,6 +8,7 @@ open class Player: AVPlayerViewController {
     static var hasAlreadyRegisteredPlaybacks = false
     private let baseObject = BaseObject()
     private var isChromeless: Bool { core?.options.bool(kChromeless) ?? false }
+    private var nextFocusEnvironment: UIFocusEnvironment?
```
fcapinho, line 11: Making this a `UIFocusEnvironment` will require sharing the actual `UICorePlugin` reference. There is no guarantee for when it will be released after adding it to an event user info.

Can we achieve the same goal (identifying the `UIView` that required the focus) without using this kind reference (e.g.: with a simple integer or string identification) ?

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -41,6 +42,14 @@ open class Player: AVPlayerViewController {
         }
     }
 
+    override open var preferredFocusEnvironments: [UIFocusEnvironment] {
+        if let view = nextFocusEnvironment as? UIView, view.alpha > 0.1 {
```
fcapinho, line 46: I consider that the `view.alpha > 0.1` check is not applicable to all `UIView`, it only works for an specific subset of `UICorePlugins`.

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
```diff
@@ -9,6 +9,16 @@ open class UICorePlugin: SimpleCorePlugin, UIPlugin {
...
+        core?.trigger(.requestFocusUpdate, userInfo: userInfo)
+    }
+    
+    func updateFocus() {
```
fcapinho, line 17: I consider that this method should not be provided: it only works if `func updateFocus(userInfo: EventUserInfo)` has already been called and requires that a strong reference to an existing `Plugin` be kept by the event receiver.

Sources/Clappr_tvOS/Classes/Base/Player.swift
pauloSantosMobileTW: Done

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
pauloSantosMobileTW: We changed it to 2 functions, 1 to request focus and other to request a update on focus engine.

Sources/Clappr_tvOS/Classes/Base/Player.swift
pauloSantosMobileTW: We create a generic abstraction of the properties that make the focus engine can not focus a view, please check it again.

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
pauloSantosMobileTW: The behavior of update focus is to trigger the update focus of the Focus Engine from tvOS and it will change the focus depending of the state of the view, we can move this for other layer if it makes more sense.

Sources/Clappr_tvOS/Classes/Base/Player.swift
paulogama14: Changes made.

Sources/Clappr/Classes/Plugin/Core/UICorePlugin.swift
pauloSantosMobileTW: Done

Sources/Clappr_tvOS/Classes/Extension/UIFocusEnvironment+ext.swift
```diff
@@ -0,0 +1,6 @@
+extension UIFocusEnvironment {
+    var isFocusable: Bool {
```
fcapinho, line 2: Which is the testing strategy for this extension ?

Sources/Clappr_tvOS/Classes/Base/Player.swift
```diff
@@ -141,6 +156,27 @@ open class Player: AVPlayerViewController {
         core?.on(Event.didChangeActivePlayback.rawValue) { [weak self] _ in self?.bindPlaybackEvents() }
         core?.on(Event.didEnterFullscreen.rawValue) { [weak self] (info: EventUserInfo) in self?.forward(.requestFullscreen, userInfo: info) }
         core?.on(Event.didExitFullscreen.rawValue) { [weak self] (info: EventUserInfo) in self?.forward(.exitFullscreen, userInfo: info) }
+        core?.on(Event.requestFocus.rawValue) { [weak self] (info: EventUserInfo) in self?.handleRequestFocusEvent(userInfo: info) }
+        core?.on(Event.releaseFocus.rawValue) { [weak self] (info: EventUserInfo) in self?.releaseFocus(userInfo: info) }
+    }
+    
+    private func handleRequestFocusEvent(userInfo: EventUserInfo) {
```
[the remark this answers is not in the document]
paulogama14: Changes made!
Commit: 7aa964440ce4d05182a20ac0d9d65164a0fb9384

Tests/Clappr_Tests/Classes/Plugin/Core/UICorePluginTests.swift
```diff
@@ -42,6 +42,43 @@ class UICorePluginTests: QuickSpec {
...
+                    }
+
+                    plugin.requestFocus()
+
+                    expect(didCallRequestFocusWithTag).toEventually(equal(123))
+                }
+            }
+
+            describe("#requestFocus") {
+                it("triggers a requestFocus event with viewTag") {
```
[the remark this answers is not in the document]
paulogama14: Changes made!
Commit: 50a46f6fd97c1aecdaa88a28327d555376fe6f5b

Sources/Clappr_tvOS/Classes/Extension/UIFocusEnvironment+ext.swift
paulogama14: Changes made!
Commit: 6c47086069f015d36a39390ba72f863f8792b7f6

PR #389, opened 2020-06-26
Replay Button

BruunoMM:
Goal
---
To create the replay feature, by showing a replay button at the end of the video, which when clicked would start the video from the beginning.

How to Test
---
1. Start the sample
1. Play the video
1. When the video ends, a replay button should show up instead of the video poster
1. Tap that button.
The video should start playing from the beginning.

### Seek & Drawer Scenarios
The media control is now avaible as the video ends, so we've had to define the button's behavior for other interactions as well, such as seeking and showing / hiding drawers.

#### Seek
1. Play a video and when it ends, the media control should show up with the replay button
1. Seek back
The replay button should turn into a play button, since you're not at the end of the video anymore
1. Play the video
1. When the video reaches the end once again, the replay should show up again normally.

#### Drawer
1. Add this chunk code on `ViewController.swift#110`:
```swift
    }

        return 32.0
    }

    override func render() {
        super.render()

        view.backgroundColor = .red
    }
}
```
2. Add this one to `ViewController.swift`, before the Player is instantiated, around `line 93`:
```swift
Player.register(plugins: [CustomBottomDrawerPlugin.self])
```

3. Run the sample and play the video
3. Open the drawer and let it open until the video ends.
The media control should not be shown.
3. Closing the drawer should show the media control with the replay button.


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -72,7 +73,8 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
             }
 
             listenTo(playback, eventName: Event.didComplete.rawValue) { [weak self] _ in
-                self?.hide()
+                guard let isDrawerActive = self?.isDrawerActive, !isDrawerActive else { return }
+                self?.show()
```
paulogama14, line 77: Maybe we can extract this logic into a method.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -26,6 +27,11 @@ open class PlayButton: MediaControl.Element {
     private var canShowPauseIcon: Bool {
         activePlayback?.state == .playing
     }
+    
+    private var shouldReplay: Bool {
+        guard let playback = activePlayback else { return false }
+        return playback.position >= playback.duration
```
paulogama14, line 33:
```suggestion

        return playback.position >= playback.duration
```

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -38,6 +44,11 @@ open class PlayButton: MediaControl.Element {
         listenTo(playback, event: .playing) { [weak self] _ in self?.onPlay() }
         listenTo(playback, event: .stalling) { [weak self] _ in self?.hide() }
         listenTo(playback, event: .didStop) { [weak self] _ in self?.onStop() }
+        listenTo(playback, event: .didComplete) { [weak self] _ in self?.onComplete() }
+        listenTo(playback, event: .willSeek) { [weak self] info in
+            guard let position = info?["position"] as? Double else { return }
+            self?.onWillSeek(position)
```
paulogama14, line 50: I think we can pass info on method.

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -71,7 +82,18 @@ open class PlayButton: MediaControl.Element {
...
+
+    private func onComplete() {
+        show()
+        changeToReplayIcon()
+    }
     
+    private func onWillSeek(_ position: Double) {
+        guard let playback = activePlayback, position != playback.duration else { return }
+        show()
+        canShowPlayIcon ? changeToPlayIcon() : changeToPauseIcon()
```
paulogama14, line 94: The method show is already present on `changeIcons`
```suggestion
        canShowPlayIcon ? changeToPlayIcon() : changeToPauseIcon()
```

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
```diff
@@ -71,7 +82,18 @@ open class PlayButton: MediaControl.Element {
         show()
         changeToPlayIcon()
     }
+
+    private func onComplete() {
+        show()
+        changeToReplayIcon()
+    }
     
+    private func onWillSeek(_ position: Double) {
```
paulogama14, line 91: Could we cover this method with more unit tests?

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
BruunoMM: I think `show` is not called inside those methods but rather on the `onPause` and `onStop` 🤔

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/PlayButton.swift
BruunoMM: Done! Refactored the whole test class.

Tests/Clappr_Tests/Classes/Plugin/Core/MediaControl/MediaControlTests.swift
```diff
@@ -315,34 +315,38 @@ class MediaControlTests: QuickSpec {
...
+                            mediaControl.show() { didCallShow = true }
+                            
+                            expect(didCallShow).to(beTrue())
+                        }
                     }
                 }
 
                 func mediaControlHidden() {
-                    coreStub.activePlayback?.trigger(Event.didComplete)
+                    mediaControl.hide()
```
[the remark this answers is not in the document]
BruunoMM: We'll address improvements for this test class in another PR.
There are other functions like this one that could be removed and improve the code's legibility, so we've decided to do this in the upcoming PR. Stay tuned hahaha 😋

PR #390, opened 2020-07-17
Refactor media control

BruunoMM:
Goal
---
To refactor `MediaControl` code and adjust its behaviour.

How to Test
---
1. Interact with the Media Control and check if it respects the following rules:

    - Every transition is animated, including showing/hiding drawers
    - It appears at the beggining of the video after the first play
    - Has a inactivity time of 3 seconds
    - It automatically disappears after 0.3 seconds after a user interaction when `playing`
    - It does not automatically disappears after user interaction when `paused` / `idle`

2. Run tests

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControlElement.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControlElement.swift
@@ -27,4 +27,10 @@ extension MediaControl {
             return .left
         }
     }
+    
+    enum AnimationState {
+        case showing
+        case hiding
+        case none
+    }
 }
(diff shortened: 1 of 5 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControlElement.swift
```diff
@@ -27,4 +27,10 @@ extension MediaControl {
             return .left
         }
     }
+
```
cairano, line 30: What do you think about moving the `enum` to the start of the escope, in the `extension` declaration, just to maintain the standard in code?

paulogama14: Nice one. Was changed on: cf9d46f7ce2fe6afe627748935d03c385504882a

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControlElement.swift
```diff
@@ -19,12 +19,15 @@ public enum MediaControlPosition {
...
+}
 
-        open var position: MediaControlPosition {
-            return .left
-        }
+extension MediaControl {
+    enum AnimationState {
+        case showing
+        case hiding
+        case none
```
cairano, line 31: Can we put the `enum` at the beginning of the escopo of the extension, just to maintain a pattern?

PR #391, opened 2020-07-22
Add Background Layer

geisonfgf:
## 🏁 Goal

Create the new layers structure and add the Background Layer to it.

## ✅ How to test

1. Run the Clappr example app
2. Open the View Hierarchy Debugger
3. Check if views follow the image below:
BackgroundLayer

### DIFF
--- a/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
@@ -1,9 +1,13 @@
-//
-//  BackgroundLayer.swift
-//  Clappr_Tests
-//
-//  Created by yves on 20/07/20.
-//  Copyright © 2020 CocoaPods. All rights reserved.
-//
+import UIKit
 
-import Foundation
+final class BackgroundLayer: UIView {
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        backgroundColor = UIColor.black
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
--- a/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
@@ -1,10 +1,10 @@
 import UIKit
 
 final class BackgroundLayer: UIView {
-    
     init() {
-        super.init(frame: CGRect.zero)
-        backgroundColor = UIColor.black
+        super.init(frame: .zero)
+
+        backgroundColor = .black
     }
     
     required init?(coder: NSCoder) {
--- a/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
+++ b/Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
@@ -1,0 +1,9 @@
+//
+//  BackgroundLayer.swift
+//  Clappr_Tests
+//
+//  Created by yves on 20/07/20.
+//  Copyright © 2020 CocoaPods. All rights reserved.
+//
+
+import Foundation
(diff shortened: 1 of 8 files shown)


Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
```diff
@@ -0,0 +1,25 @@
+import UIKit
+
+final class BackgroundLayer: UIView {
+    
+    init() {
+        super.init(frame: CGRect.zero)
+        backgroundColor = UIColor.black
```
paulogama14, line 7:
```suggestion
    init() {
        super.init(frame: .zero)

        backgroundColor = .black
```

Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
```diff
@@ -0,0 +1,25 @@
...
+    func attach(to view: UIView) {
+        view.addSubview(self)
+        
+        translatesAutoresizingMaskIntoConstraints = false
+        
+        widthAnchor.constraint(equalTo: view.widthAnchor).isActive = true
+        heightAnchor.constraint(equalTo: view.heightAnchor).isActive = true
+        
+        centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+        centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
```
paulogama14, line 23: Maybe we can replace this code with the `UIView+Ext`, with the `anchorInCenter` and `setWidthAndHeight` methods. What do you think?

Sources/Clappr_iOS/Classes/Base/Layers/LayersCompositor.swift
```diff
@@ -0,0 +1,15 @@
...
+
+class LayersCompositor {
+    private weak var rootView: UIView?
+    private var backgroundLayer: BackgroundLayer?
+    
+    func compose(inside rootView: UIView) {
+        self.rootView = rootView
+        
+        let backgroundLayer = BackgroundLayer()
+        self.backgroundLayer = backgroundLayer
```
paulogama14, line 11: I think this `let backgroundLayer` doesn't need to be instantiated and the fact to create an instance of `BackgroundLayer` inside the method could show this method have more than one responsibility. But removing this, we have to refactor the `backgroundLayer` and the class. What do you think about it?

Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
torives: In this case, `anchorInCenter` works great. But `setWidthAndHeight` would make the layer size constant, but what we actually need here is for it to match its container (the Core view) size. So we kept the constraints definition together instead of using only one extension method, as we thought it makes the code more readable.

We also agreed that this extension needs a little bit of cleaning up. It's kinda messy right now with methods with multiple responsibilities and using the visual constraint language, just to name a few issues with it. The initial idea was to create a new PR with these improvements and then update this code.

Sources/Clappr_iOS/Classes/Base/Layers/LayersCompositor.swift
torives: What is your suggestion? To initialize `backgroundLayer` upon declaration to get rid of the optional value? If so, I do like this idea better.

Sources/Clappr_iOS/Classes/Base/Layers/BackgroundLayer.swift
paulogama14: Nice. I agree with you
=]

Sources/Clappr_iOS/Classes/Base/Layers/LayersCompositor.swift
paulogama14: Yeap
=)

torives: Done in 3b71f1ef5d439bcfe14b4aebbe1a4721159d669e

PR #392, opened 2020-08-03
Subtitle Style

cairano:
# Goal

Introduce player capability to change the subtitle style.
Create an abstraction from `AVTextStyleRules` to use a non-verbose code

# Class Diagram

UML

# How to Test

Change the `options` on `DashboardViewController.swift#19`

```swift
let options: Options = [
            kSourceUrl: "http://sample.vodobox.com/planete_interdite/planete_interdite_alternate.m3u8",
            kPosterUrl: "http://clappr.io/poster.png",
            kDefaultSubtitle: "eng",
            kFullscreen: switchFullscreen.isOn,
            kFullscreenByApp: switchFullscreenControledByApp.isOn
        ]
```

Change the `selectDefaultMediaOptionIfNeeded` function  on `AVFoundationPlayback.swift#650`

```swift
func selectDefaultMediaOptionIfNeeded() {
        selectDefaultAudioIfNeeded()
        changeSubtitle(style: [ 
            .font(.fantasy), 
            .bold, 
            .background(.blue) 
        ])
        selectDefaultSubtitleIfNeeded()
    }
```

Run `Clappr_Example`

### Expected Result (iOS)
Simulator Screen Shot - iPhone Xs - 2020-08-05 at 16 48 02

### Expected Result (iPadOS)
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-08-05 at 17 12 11

Run `Clappr_tvOS_Example`

### Expected Result (tvOS)
Simulator Screen Shot - Apple TV 4K (at 1080p) - 2020-08-05 at 17 06 53

### DIFF
--- a/Sources/Clappr/Classes/Base/Playback.swift
+++ b/Sources/Clappr/Classes/Base/Playback.swift
@@ -106,7 +106,6 @@ open class Playback: UIObject, NamedType {
     @objc open func seek(_: TimeInterval) {}
     @objc open func seekToLivePosition() {}
     open func mute(_: Bool) {}
-    open func changeSubtitle(style textStyle: [TextStyle]) {}
 
     @objc open func destroy() {
         Logger.logDebug("destroying", scope: "Playback")
--- a/Sources/Clappr/Classes/Base/Playback.swift
+++ b/Sources/Clappr/Classes/Base/Playback.swift
@@ -106,6 +106,7 @@ open class Playback: UIObject, NamedType {
     @objc open func seek(_: TimeInterval) {}
     @objc open func seekToLivePosition() {}
     open func mute(_: Bool) {}
+    open func changeSubtitle(style textStyle: [TextStyle]) {}
 
     @objc open func destroy() {
         Logger.logDebug("destroying", scope: "Playback")
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -635,7 +635,7 @@ open class AVFoundationPlayback: Playback {
         #endif
     }
 
-    public func changeSubtitle(style textStyle: [TextStyle]) {
+    open override func changeSubtitle(style textStyle: [TextStyle]) {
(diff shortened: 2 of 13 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -759,4 +759,8 @@ open class AVFoundationPlayback: Playback {
             view.accessibilityIdentifier = "AVFoundationPlaybackError"
         }
     }
+
+    public func changeSubtitle(style textStyle: [TextStyle]) {
```
paulogama14, line 763: I think we can put this code near the other subtitle methods. What do you think?

paulogama14: Another thing, I think we can increment the unit tests to cover this part of code.
=]

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -635,6 +635,10 @@ open class AVFoundationPlayback: Playback {
         #endif
     }
 
+    open override func changeSubtitle(style textStyle: [TextStyle]) {
+        self.player?.currentItem?.textStyle = textStyle
```
geisonfgf, line 639: This method has a currentItem nullable as a dependency and it is possible to be called without a currentItem available, this way this method will do nothing silently what could cause a semantical problem. Maybe a possible solution would be returns a bool or an exception when the dependency was not attended.

Sources/Clappr/Classes/Base/Playback.swift
```diff
@@ -106,6 +106,7 @@ open class Playback: UIObject, NamedType {
     @objc open func seek(_: TimeInterval) {}
     @objc open func seekToLivePosition() {}
     open func mute(_: Bool) {}
+    open func changeSubtitle(style textStyle: [TextStyle]) {}
```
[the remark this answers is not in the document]
helbertgs: Ok

PR #396, opened 2020-08-28
Create Container & Playback layers

torives:
## 🏁 Goal
Continue the implementation of the visual layers abstraction by adding Container and Playback layers.

## ✅ How to test
1. Add the following snippet to LayerComposer.swift#13
```swift
playbackLayer.backgroundColor = .yellow
containerLayer.backgroundColor = .red
```
2. Run the Clappr_Example project
3. Enable view hierarchy debugger
4. You should be able to see the new layers

## 🖼 Screenshots
Screen Shot 2020-08-28 at 14 10 59


Tests/Clappr_Tests/Classes/Base/Layers/ContainerLayerTests.swift
```diff
@@ -0,0 +1,39 @@
...
+                    let containerView = UIView(frame: .zero)
+                    let frame = CGRect(x: 0, y: 0, width: 100, height: 100)
+                    let layer = ContainerLayer(frame: frame)
+                    
+                    layer.attachContainer(containerView)
+                    layer.layoutIfNeeded()
+                    
+                    expect(containerView.center).to(equal(layer.center))
+                    expect(containerView.bounds.size).to(equal(layer.bounds.size))
+
```
paulogama14, line 19: I think that we can change for explicit values and not related objects inside the test arrange, once we can change the test arrange or the ContainerLayer code and the code will not reflect the changes or cover the main behavior. What do you think about it?

Tests/Clappr_Tests/Classes/Base/Layers/ContainerLayerTests.swift
```diff
@@ -0,0 +1,39 @@
...
+                        let bigFrame = CGRect(x: 0, y: 0, width: 100, height: 100)
+                        let containerView = UIView(frame: .zero)
+                        let containerLayer = ContainerLayer(frame: smallFrame)
+                        
+                        containerLayer.attachContainer(containerView)
+                        containerLayer.layoutIfNeeded()
+                        containerLayer.bounds = bigFrame
+                        containerLayer.layoutIfNeeded()
+                        
+                        expect(containerView.bounds.size).to(equal(bigFrame.size))
```
paulogama14, line 33: Same here.

Sources/Clappr/Classes/Base/Layers/LayerComposer.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+
+public class LayerComposer {
+    private weak var rootView: UIView?
+    private let backgroundLayer = BackgroundLayer()
+    private let playbackLayer = PlaybackLayer()
+    private let containerLayer = ContainerLayer()
+    
+    public init(){}
```
paulogama14, line 9: Maybe we can omit this init.

```suggestion
```

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -107,11 +106,18 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
     
     open func attach(to parentView: UIView, controller: UIViewController) {
+        guard let containerView = activeContainer?.view else {
+            Logger.logError("Active container should not be nil when attaching to parent view", scope: "Core")
+            return
+        }
```
paulogama14, line 112: Is there a direct way to cover this scenario with unit test?

Sources/Clappr/Classes/Base/Layers/LayerComposer.swift
torives: As LayerComposer is a parameter to Core and Container required initializers, and those initializers are public, we need to declare it as a public class also. And as such, we need to enable the consumers of our lib to instantiate it, or else they won't be able to use Core or Container's initializer also.

Sources/Clappr/Classes/Base/Core.swift
torives: This scenario is an aberration. Meaning that although it is possible in the compiler context, on the application context we **always** initialize a Core with a Container at the CoreFactory. IMO, `activeContainer` as an optional property shouldn't even exist.

Having said that, I don't think there's an actual gain in testing it, as it won't ever happen 🤷🏽‍♂️

Sources/Clappr/Classes/Base/Layers/LayerComposer.swift
paulogama14: Got it.
;)

Sources/Clappr/Classes/Base/Core.swift
paulogama14: I tend to think about tests in a different way. As much as the responsibility for creating these objects lies with other entities, it is the responsibility of Core to ensure that these objects exist. He doesn't know the moment they are created. Another point is that where these objects are created they can be modified and the `Core` can no longer have this security and break because there is no `activeContainer` at that moment.
So I see the creation of this test as a gain.
Makes sense?

paulogama14 (approved):

Sources/Clappr/Classes/Base/Core.swift
torives: It really does. We added the test on 1926bc87bf79ed653cb3d5860f028d398884aa4a

Tests/Clappr_Tests/Classes/Base/Layers/ContainerLayerTests.swift
torives: Solved on 9614ccf02307e4d9d30380cdaf81a0471999f494

Tests/Clappr_Tests/Classes/Base/Layers/ContainerLayerTests.swift
torives: Solved on 9614ccf02307e4d9d30380cdaf81a0471999f494

PR #398, opened 2020-09-01
Read offline assets from assetCache

lfonseca:
## 🏁  Goal
Read only the available tracks from offline assets to prevent network access during the offline experience.

## ✅  How to test
1. Embed any offline asset in the application (e.g. the sample.movpkg in the test bundle).
2. Turn off the network access.
3. Make sure all the tracks being shown are available offline.


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -119,13 +119,24 @@ open class AVFoundationPlayback: Playback {
...
+            availableOptions = mediaGroup.options.compactMap({ MediaOptionFactory.fromAVMediaOption($0, type: .subtitle) })
+        }
         return availableOptions + [MediaOption.offSubtitle]
     }
 
     open override var audioSources: [MediaOption]? {
         guard let mediaGroup = mediaSelectionGroup(.audible) else { return [] }
-        return mediaGroup.options.compactMap({ MediaOptionFactory.fromAVMediaOption($0, type: .audioSource) })
+        let availableOptions: [MediaOption]
+        if let cache = asset?.assetCache, cache.isPlayableOffline {
```
johnmartinstw, line 134: What if we created a function encapsulating this logic of capturing the `MediaOption` list? We could pass the media type (`.audioSource` or `.subtitle`) and avoid this code repetition. WDYT?

lfonseca: Done!

johnmartinstw (approved):
Nice job, guys! 🚀

dezouzart (approved):

PR #399, opened 2020-09-02
Fix bottom drawer gesture handler

VitorFerraz:
## 🏁  Goal
Fix gesture handler on drag bottom drawer when showing and hiding 

## ✅  How to test
1. Run unit tests.
2. Interact with drawer

On `ViewController.swift` insert:

`Line #94`
```swift
Player.register(plugins: [CustomBottomDrawerPlugin.self])
```

`Line #110`
```swift
    }
        return 32.0
    }
    override func render() {
        super.render()
        view.backgroundColor = .red
    }
}
```

### DIFF
--- a/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
+++ b/Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
@@ -28,8 +28,8 @@ open class BottomDrawerPlugin: DrawerPlugin {
     required public init(context: UIObject) {
         super.init(context: context)
 
-        addGesture(UITapGestureRecognizer(target: self, action: #selector(didTapView)))
-        addGesture(UIPanGestureRecognizer(target: self, action: #selector(onDragView)), cancelsTouchesInView: true)
+        addGesture(UITapGestureRecognizer(target: self, action: #selector(didTapView)), cancelingTouchesInView: false)
+        addGesture(UIPanGestureRecognizer(target: self, action: #selector(onDragView)))
     }
 
     override open func render() {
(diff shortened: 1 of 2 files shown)


Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -57,15 +57,21 @@ open class BottomDrawerPlugin: DrawerPlugin {
...
     }
 
     private func moveDown(with duration: TimeInterval = ClapprAnimationDuration.mediaControlHide) {
+        enableUserInteractionOnFirstSubview(false)
         view.setVerticalPoint(to: hiddenHeight, duration: duration)
     }
 
-    private func addGesture(_ gesture: UIGestureRecognizer) {
-        gesture.cancelsTouchesInView = false
+    private func enableUserInteractionOnFirstSubview(_ enabled: Bool) {
```
torives, line 69:
```suggestion
    private func toggleContentInteraction(enabled: Bool) {
```

I find kinda odd having to pass a boolean to a method called "enable X" in order for it to actually enable something. It's odder still passing `false` to it. Calling it `toggle` and exposing the parameter name kinda solves that, making intention clearer IMHO.

And using "content" instead of "subview" is just to leave it closer to the real objective, instead of how it's actually done.

Sources/Clappr_iOS/Classes/Plugin/Core/Drawer/BottomDrawerPlugin.swift
```diff
@@ -57,15 +57,21 @@ open class BottomDrawerPlugin: DrawerPlugin {
...
+        self.view.subviews.first?.isUserInteractionEnabled = enabled
+    }
+
+    private func addGesture(_ gesture: UIGestureRecognizer, cancelsTouchesInView: Bool = false) {
```
torives, line 73:
```suggestion
    private func addGesture(_ gesture: UIGestureRecognizer, cancelingTouchesInView: Bool = true) {
```

Considering that `cancelsTouchesInView` default value is true on UIGestureRecognizers, I think we should invert the default parameter to also be true. This would avoid confusion by making clear we're altering the default behavior on the call site.

Also, just changed the verb tense so it reads better. What are your thoughts on this?

VitorFerraz: That makes sense to me! 
👍

torives (approved):

helbertgs (approved):

PR #404, opened 2020-09-14
Disable external playback

torives:
## 🏁 Goal
Allow the client application to disable external playback, such as AirPlay.

## ✅ How to test
⚠️  As there's no AirPlay support on the simulator, to test this you'll need a device to run the example application and another one to use as the AirPlay destination, such as an Apple TV.

### Scenario 1: AirPlay enabled
1. Run the Example app
2. Play the video
3. Open the Control Center
4. Connect to the Apple TV
5. Video and audio should play normally on the TV

### Scenario 2: Airplay disabled
1. On the example app, add the following line at ViewController.swift:94
```swift
options[kDisableExternalPlayback] = true
```
1. Build and run the Example app
2. Play the video
3. Open the Control Center
4. Connect to the Apple TV
5. Video and audio should play locally

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -467,7 +467,7 @@ open class AVFoundationPlayback: Playback {
     
     private func setAudioSessionCategory(to category: AVAudioSession.Category, with options: AVAudioSession.CategoryOptions = []) {
         do {
-            try AVAudioSession.sharedInstance().setCategory(category, mode: .default, options: options)
+            try AVAudioSession.sharedInstance().setCategory(category, mode: .moviePlayback, options: options)
         } catch {
             Logger.logError("It was not possible to set the audio session category")
         }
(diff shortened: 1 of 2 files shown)


Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -260,7 +260,8 @@ open class AVFoundationPlayback: Playback {
         if let player = player as? AVQueuePlayer {
             playerLooper = AVPlayerLooper(player: player, templateItem: item)
         }
-        player?.allowsExternalPlayback = true
+        
+        player?.allowsExternalPlayback = !options.bool(kDisableExternalPlayback, orElse: false)
```
paulogama14, line 264: In my opinion, reading was hindered by the number of possible cases in this variable. Could we extract into a more expressive one, like an external var?
What do you think?

torives: Sure can! Done in bbe83474317c6ec466c521d7406e84cc38352535

geisonfgf (approved):

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
paulogama14: =]

paulogama14 (approved):

helbertgs:
Scenario 1 isn't working

torives:
> Scenario 1 isn't working

Could you elaborate a bit? What went wrong? Did everything work as expected in the second scenario?

helbertgs:
In scenario 1, the audio should go to Apple TV, and this behavior did not occur. 
In scenario 2 apparently worked correctly.

PR #405, opened 2020-09-22
Stop Playback before destroying it

torives:
## 🏁 Goal
Stop playback before destroying the player in order to maintain state integrity.

## ✅ How to test
- Run the automated tests

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -523,6 +523,8 @@ open class AVFoundationPlayback: Playback {
     }
 
     open override func stop() {
+        guard state != .idle else { return }
+        
         isStopped = true
         trigger(.willStop)
         updateState(.idle)
--- a/Tests/Clappr_Tests/Classes/Base/PlaybackTests.swift
+++ b/Tests/Clappr_Tests/Classes/Base/PlaybackTests.swift
@@ -87,6 +87,14 @@ class PlaybackTests: QuickSpec {
 
                 expect(callbackWasCalled) == false
             }
+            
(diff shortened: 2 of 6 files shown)


geisonfgf (approved):

paulogama14 (approved):

Tests/Clappr_Tests/Classes/Base/PlaybackTests.swift
```diff
@@ -87,6 +87,14 @@ class PlaybackTests: QuickSpec {
...
                 expect(callbackWasCalled) == false
             }
+            
+            it("stops playing when destroy is called") {
+                let playbackSpy = PlaybackSpy(options: options)
+                
+                playbackSpy.destroy()
+                
+                expect(playbackSpy.didCallStop).to(beTrue())
+            }
```
paulogama14, line 97: I think we can change the tests for the concrete entity. The new scenario (check if isStopped is false) was not covered by tests and the internal changes of `AVFoundationPlayback.stop()` will not reflect the reality on `PlaybackSpy`. Maybe we can change the expectation for one of the triggered events inside stop.
What do you think?

paulogama14 (approved):

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -522,7 +522,7 @@ open class AVFoundationPlayback: Playback {
         updateState(.paused)
     }
 
-    open override func stop() {
+    open override func stop() {
```
paulogama14, line 525:
```suggestion
    open override func stop() {
```

paulogama14 (approved):

geisonfgf (approved):

paulogama14 (approved):

PR #406, opened 2020-09-28
Create MediaControl layer

helbertgs:
## 🏁 Goal
Continue the implementation of the visual layers abstraction by adding Media Control layer.

## ✅ How to test
1. Add the following snippet to LayerComposer.swift#19
```swift
mediaControlLayer.backgroundColor = .yellow
```
2. Run the Clappr_Example project
3. Enable view hierarchy debugger
4. You should be able to see the new layer

## 🖼 Screenshots
Screen Shot 2020-09-28 at 16 14 19

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -175,8 +175,8 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         guard let mediaControl = plugins.first(where: { $0 is MediaControl }) as? MediaControl else {
             return
         }
-        let mediaControlView = mediaControl.view
-        layerComposer.attachMediaControl(mediaControlView)
+
+        layerComposer.attachMediaControl(mediaControl.view)
         mediaControl.safeRender()
     }
 
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -109,22 +109,12 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         parentView.addSubviewMatchingConstraints(view)
     
         layerComposer.attachContainer(containerView)
-        setupMediaControlLayer()
         layerComposer.compose(inside: view)
         
         self.parentController = controller
         self.parentView = parentView
     
         trigger(.didAttachView)
-    }
-    
-    private func setupMediaControlLayer() {
-        #if os(iOS)
-        let mediaControl = plugins.first { $0 is MediaControl } as? MediaControl
-        if let mediaControlView = mediaControl?.view {
-            layerComposer.attachMediaControl(mediaControlView)
-        }
-        #endif
     }
     
     open override func render() {
@@ -171,10 +161,19 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
 
     #if os(iOS)
     private func renderCorePlugins() {
-        plugins
-            .filter { $0.isNotMediaControlElement }
-            .filter { $0.isNotOverlayPlugin}
-            .forEach(render)
+        plugins.filter { plugin in
+            switch plugin {
+            case let (mediaControl as MediaControl):
+                layerComposer.attachMediaControl(mediaControl.view)
+                render(mediaControl)
+                return false
+            case is OverlayPlugin, is MediaControl.Element:
+                return false
+            default:
+                return true
+            }
+        }
+        .forEach(render)
     }
 
     private func renderMediaControlElements() {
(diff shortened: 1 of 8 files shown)


Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -163,11 +163,22 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
+            .filter { $0.isNotMediaControl }
             .forEach(render)
     }
+    
+    private func attachMediaControlLayer() {
+        guard let mediaControl = plugins.first(where: { $0 is MediaControl }) as? MediaControl else {
+            return
+        }
+        let mediaControlView = mediaControl.view
+        layerComposer.attachMediaControl(mediaControlView)
```
paulogama14, line 179:
```suggestion
        }

        layerComposer.attachMediaControl(mediaControl.view)
```

paulogama14 (approved):

torives (approved):

Tests/Clappr_Tests/Classes/Plugin/Core/CoreTests.swift
```diff
@@ -154,6 +159,22 @@ class CoreTests: QuickSpec {
                             expect(didAttachPlayer).to(beFalse())
                         }
                     }
+
+                    #if os(iOS)
+                    context("when calls attach") {
+                        it("should calls the attachMediaControl on LayerComposer") {
```
BruunoMM, line 165:
```suggestion
                    context("when attach is called") {
                        it("should attach the MediaControl layer on LayerComposer") {
```

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
```diff
@@ -0,0 +1,39 @@
...
+    override func spec() {
+        describe(".MediaControlLayer") {
+            context("When a MediaControl is attached") {
+                it("resizes to match the layer bounds") {
+                    let mediaControlView = UIView(frame: .zero)
+                    let frame = CGRect(x: 0, y: 0, width: 100, height: 100)
+                    let layer = MediaControlLayer(frame: frame)
+
+                    layer.attachMediaControl(mediaControlView)
+                    layer.layoutIfNeeded()
```
BruunoMM, line 15: Do we need this call? If we do, could we encapsulate it inside the attach method itself so it's contained?

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
```diff
@@ -0,0 +1,39 @@
...
+                }
+                context("and the view size changes") {
+                    it("resizes to match the view bounds") {
+                        let smallFrame = CGRect(x: 0, y: 0, width: 50, height: 50)
+                        let bigFrame = CGRect(x: 0, y: 0, width: 100, height: 100)
+                        let mediaControlView = UIView(frame: .zero)
+                        let mediaControlLayer = MediaControlLayer(frame: smallFrame)
+
+                        mediaControlLayer.attachMediaControl(mediaControlView)
+                        mediaControlLayer.layoutIfNeeded()
```
BruunoMM, line 29: Same as above. Let me know what you think about the suggestion.

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
torives: This call is not needed, as the implementation code already calls it.

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
torives: This call is as redundant as the one on the above comment.

But just to offer some context, these calls to `layoutIfNeeded` are necessary to force the view size update. 

We're calling `translatesAutoresizingMaskIntoConstraints = false` in the implementation code to avoid conflicts in our constraints. But this actually invalidates direct frame modifications, so during testing, we need to call `layoutIfNeeded` to make AutoLayout do its magic.

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
VitorFerraz: Yes, we do, but it not necessary to encapsulate this inside the attach method because the `LayerComposer` already handles this when we call the `compose`. Since we are testing only the `MediaControlLayer` we need to explicitly call  `layoutIfNeeded`

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
VitorFerraz: Same answer as above.

cairano (approved):

Tests/Clappr_Tests/Classes/Base/Layers/MediaControlLayerTests.swift
VitorFerraz: We changed the media control layer to update its parent view layout instead of MediaControlView (child view) on `attachMediaControl(_ mediaControl: UIView)`

BruunoMM (approved):
🚀

PR #412, opened 2020-10-22
Changes the access control on core

cairano:
Goal
---
Expose access control for `Core` property on Player to allow external frameworks to be able to access and listen to `Core` events on the correct context layer.

How To Test
---
Classes on external frameworks that inherited from `Player.swift` should be able to access the property `core` only as a getter.


helbertgs (approved):

BruunoMM (approved):

VitorFerraz:
:shipit:

PR #417, opened 2020-11-24
Chromeless mode for iOS

VitorFerraz:
## Goal

Introduce `Chromeless` mode to remove user interaction and visual elements on iOS.

## How to test
Scenario 1
1 - Add the code on ViewController.swift#18:
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
    self.player.chromelessMode = true
}

DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
    self.player.chromelessMode = false
}
```
- Run the playground
- At 00:05s the interactions with playback will be disabled and the visual elements will be hide.
- At 00:10s the interactions with playback will be enabled.

Scenario 2
- Add the code on ViewController.swift#18:
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
    self.player.configure(options: [kChromeless: true])
}

DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
    self.player.configure(options: [kChromeless: false])
}
```
- Run the playground
- At 00:05s the interactions with playback will be disabled and the visual elements will be hide.
- At 00:10s the interactions with playback will be enabled.

Scenario 3
Add the code on ViewController.swift#94:
```swift
options[kChromeless] = true
```
- Run the playground
- The interactions with playback will be disabled and the visual elements will be hidden.

## 🖼  Screenshot
image

### DIFF
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -69,7 +69,8 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         self.options = options
         self.layerComposer = layerComposer
         super.init()
-        addTapGestures()
+        
+        updateChromelessMode()
         
         bindEventListeners()
         
@@ -151,9 +152,10 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
     
     open override func render() {
+        addTapGestures()
         containers.forEach(renderContainer)
         addToContainer()
-        updateChromelessMode()
+        updateGestureRecognizersState()
     }
 
     private func addToContainer() {
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -114,13 +114,13 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         }
     }
     
-    public func enterChromelessMode() {
+    private func enterChromelessMode() {
         #if os(iOS)
         layerComposer.hideUI()
         #endif
     }
 
-    public func exitChromelessMode() {
+    private func exitChromelessMode() {
         #if os(iOS)
         layerComposer.showUI()
         #endif
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -69,11 +69,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         self.options = options
         self.layerComposer = layerComposer
         super.init()
-        self.updateChromelessMode()
-        
-        if !chromelessMode {
-            addTapGestures()
-        }
+        addTapGestures()
         
         bindEventListeners()
         
@@ -105,7 +101,10 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
     
     private func updateGestureRecognizersState() {
+        #if os(iOS)
         view.gestureRecognizers?.forEach { $0.isEnabled = !chromelessMode }
+        #endif
+
     }
     
     private func updateChromelessMode() {
@@ -154,6 +153,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     open override func render() {
         containers.forEach(renderContainer)
         addToContainer()
+        updateChromelessMode()
     }
 
     private func addToContainer() {
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -55,7 +55,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     }
 
     @objc open var isFullscreen: Bool = false
-    private var isChromeless: Bool { options.bool(kChromeless) }
+    public var chromelessMode: Bool = false
     
     public required init(options: Options = [:], layerComposer: LayerComposer) {
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
@@ -63,8 +63,9 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         self.options = options
         self.layerComposer = layerComposer
         super.init()
-        
-        if !isChromeless {
+        self.updateChromelessMode(with: options)
+
+        if !chromelessMode {
             addTapGestures()
         }
         
@@ -91,6 +92,24 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
         if activeContainer != container {
             activeContainer = container
         }
+    }
+    
+    public func updateChromelessMode(with option: Options) {
+        if let chromelessMode = options[kChromeless] as? Bool {
+            self.chromelessMode = chromelessMode
+        }
+    }
+    
+    public func enterChromelessMode() {
+        #if os(iOS)
+        layerComposer.hideViews()
+        #endif
+    }
+
+    public func exitChromelessMode() {
+        #if os(iOS)
+        layerComposer.showViews()
+        #endif
     }
 
     private func bindEventListeners() {
--- a/Sources/Clappr/Classes/Base/Core.swift
+++ b/Sources/Clappr/Classes/Base/Core.swift
@@ -59,6 +59,7 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
     public var chromelessMode: Bool = false {
         didSet {
             updateChromelessModeVisibility()
+            updateGestureRecognizersState()
         }
     }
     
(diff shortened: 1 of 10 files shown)


Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -92,6 +93,24 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
             activeContainer = container
         }
     }
+    
+    public func updateChromelessMode(with option: Options) {
```
torives, line 97: I have a few considerations about this method.

I don't think it should be public. It kinda allows for an external caller to pass a different set of `Options` to `Core`, other than the ones received on the initializer or directly modified using the `core.options` stored property. 

Also, it seems to me that it is called in the wrong places. It seems to me that the actual trigger to this change is the modification of the `core.options` property. As we already have a `didSet` callback for it, we could call it there or better yet, **inline it**.

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -92,6 +93,24 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
+        if let chromelessMode = options[kChromeless] as? Bool {
```
torives, line 98: There's propably a typo here, and you used the stored property `options` instead of the parameter `option`, which I guess was the original intention. 

But complementing my comment on the previous line, I think we should be using the stored property 😉

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -55,16 +55,17 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
 
         self.options = options
         self.layerComposer = layerComposer
         super.init()
-        
-        if !isChromeless {
+        self.updateChromelessMode(with: options)
+
+        if !chromelessMode {
```
torives, line 68: I guess what we actually want to do here is to enable/disable de gesture recognizers based on the Chromeless state. The gestures should always be added on initialization, we just want to disable it at a particular moment.

Sources/Clappr/Classes/Base/Layers/LayerComposer.swift
```diff
@@ -35,4 +35,16 @@ public class LayerComposer {
...
+        mediaControlLayer.isHidden = false
+        coreLayer.isHidden = false
+        overlayLayer.isHidden = false
+    }
+
+    func hideViews() {
+        mediaControlLayer.isHidden = true
+        coreLayer.isHidden = true
+        overlayLayer.isHidden = true
+    }
```
torives, line 49: I suggest using the more descriptive names: `showUI` and `hideUI`. There are no parameters to these methods, so its hard to now which views are we talking about. As we are only hiding views with user interaction, I guess the `UI` is more appropriate.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -124,12 +130,28 @@ open class Player: BaseObject {
...
     
     @objc private func willEnterForeground() {
-        if isChromeless {
+        if chromelessMode {
             Logger.logDebug("forced play after return from background", scope: "Player")
             play()
         }
     }
     
+    private func updateChromelessModeVisibility() {
```
torives, line 139: As the real Chromeless state is stored on `Core`, this code should probably be there. It's called as a result of changing the `Core.chromelessMode` property. IMHO this is clear as the methods which are actually doing something are `Core` methods.

Sources/Clappr_iOS/Classes/Base/Player.swift
```diff
@@ -87,7 +92,8 @@ open class Player: BaseObject {
         bindMediaControlEvents()
         bindPlaybackEvents()
         bindNotifications()
-
+        core?.updateChromelessMode(with: options)
+        updateChromelessModeVisibility()
```
torives, line 96: If we move the `updateChromelessModeVisibility()` to `Core` as a `didSet` callback for example, there's actually no need for these methods to exist and be called during `init` or `configure`.

Sources/Clappr_iOS/Classes/Plugin/Container/PosterPlugin.swift
```diff
@@ -56,6 +51,10 @@ open class PosterPlugin: OverlayPlugin {
         view.addSubviewMatchingConstraints(poster)
 
         view.addSubview(playButton)
+
+        let xCenterConstraint = NSLayoutConstraint(item: playButton, attribute: .centerX,
```
torives, line 55: Why are we altering `PosterPlugin` constraints on this seemingly unrelated PR?

Sources/Clappr_iOS/Classes/Plugin/Core/QuickSeekCorePlugin.swift
```diff
@@ -23,6 +24,10 @@ public class QuickSeekCorePlugin: QuickSeekPlugin {
     }
     
     override func shouldSeek(point: CGPoint) -> Bool {
+        if chromelessMode {
```
torives, line 27: If we disable the gesture recognizers on `Core` (and therefore avoid triggering `didTappedCore` event), do we need this check here?

Sources/Clappr/Classes/Base/Core.swift
VitorFerraz: Actually on the init the `options` didSet is not called. 
Reference:  Swift Documentation 
[_willSet and didSet observers are not called when a property is first initialized. They are only called when the property’s value is set outside of an initialization context._](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#//apple_ref/doc/uid/TP40014097-CH14-XID_368)

Sources/Clappr_iOS/Classes/Base/Player.swift
VitorFerraz: Actually on the init the `options` didSet is not called. 
Reference:  Swift Documentation 
[_willSet and didSet observers are not called when a property is first initialized. They are only called when the property’s value is set outside of an initialization context._](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#//apple_ref/doc/uid/TP40014097-CH14-XID_368)

Sources/Clappr_iOS/Classes/Plugin/Core/QuickSeekCorePlugin.swift
VitorFerraz: I think a possible approach is to disable the `gestureRecognizers `on the core`s view because the QuickSeekCorePlugin creates another Gesture `doubleTapGesture` the adds to the core.
```
private func updateGestureRecognizers() {
        view.gestureRecognizers?.forEach{ $0.isEnabled = !chromelessMode }
}
```

Sources/Clappr/Classes/Base/Core.swift
```diff
@@ -55,18 +56,20 @@ open class Core: UIObject, UIGestureRecognizerDelegate {
...
         Logger.logDebug("loading with \(options)", scope: "\(type(of: self))")
 
         self.options = options
         self.layerComposer = layerComposer
         super.init()
-        
-        if !isChromeless {
-            addTapGestures()
-        }
+        addTapGestures()
```
torives, line 72: It worries me the fact that setting the options (when `kChromeless` is included) affects the way we handle the gestures. As of now, we're updating gestures' state before we actually add them as a result of the `didSet` chain triggered after setting the options. It seems to me that's the reason you're calling `updateChromelessMode` on `render`.

It seems to me that a better option is to actually create the gestures during `render` (which makes sense its purpose is to setup views/view related stuff), and enable/disable the gestures after its creation based on the `chomelessMode` state, which will certainly be set by this point.

Sources/Clappr_iOS/Classes/Base/Player.swift
torives: Completely forgot that. Thanks 😉

Sources/Clappr_iOS/Classes/Plugin/Core/MediaControl/MediaControl.swift
```diff
@@ -26,8 +25,7 @@ open class MediaControl: UICorePlugin, UIGestureRecognizerDelegate {
     }
     
     private var canHide: Bool { animationState != .hiding && !alwaysVisible }
-    private var canShow: Bool { !isChromeless
-        && animationState != .showing
+    private var canShow: Bool { animationState != .showing
```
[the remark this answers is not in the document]
VitorFerraz: One or two checks could be on the same line but three or more could be broken to improve readability ihmo.

torives (approved):

geisonfgf (approved):

PR #422, opened 2021-01-07
Create AVPlayer during AVFoundationPlayback's initialisation

danielbarboza:
## Goal
Move AVPlayer creation to AVFoundationPlayback initialization.

## How to test 
1 . Run all automated tests

### DIFF
--- a/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
+++ b/Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
@@ -11,7 +11,7 @@ open class AVFoundationPlayback: Playback {
     private(set) var seekToTimeWhenReadyToPlay: TimeInterval?
     private var selectedCharacteristics: [AVMediaCharacteristic] = []
 
-    @objc dynamic var player: AVPlayer?
+    @objc dynamic var player: AVPlayer!
     @objc dynamic private var playerLooper: AVPlayerLooper? {
         didSet {
             loopObserver = observe(\.playerLooper?.loopCount) { [weak self] _, _ in
@@ -19,7 +19,7 @@ open class AVFoundationPlayback: Playback {
             }
         }
     }
-    private var playerLayer: AVPlayerLayer?
+    private var playerLayer: AVPlayerLayer!
     private var playerStatus: AVPlayerItem.Status = .unknown
     private var isStopped = false
     private var timeObserver: Any?
@@ -35,7 +35,7 @@ open class AVFoundationPlayback: Playback {
 
     private var observers = [NSKeyValueObservation]()
 
-    private var lastLogEvent: AVPlayerItemAccessLogEvent? { player?.currentItem?.accessLog()?.events.last }
+    private var lastLogEvent: AVPlayerItemAccessLogEvent? { player.currentItem?.accessLog()?.events.last }
     private var numberOfDroppedVideoFrames: Int? { lastLogEvent?.numberOfDroppedVideoFrames }
     
     open var bitrate: Double? { lastLogEvent?.indicatedBitrate }
@@ -140,7 +140,7 @@ open class AVFoundationPlayback: Playback {
 
     open override var duration: Double {
         var durationTime: Double = 0
-        guard let assetDuration = player?.currentItem?.asset.duration else { return durationTime }
+        guard let assetDuration = player.currentItem?.asset.duration else { return durationTime }
 
         if playbackType == .vod {
             durationTime = CMTimeGetSeconds(assetDuration)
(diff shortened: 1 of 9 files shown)


geisonfgf (approved):

Tests/Clappr_Tests/Classes/Plugin/Playback/AVFoundationPlaybackQualityMetricsTests.swift
```diff
@@ -13,6 +13,7 @@ class AVFoundationPlaybackQualityMetricsTests: QuickSpec {
                         let baseObject = BaseObject()
                         var bitrate: Double?
                         let avfoundationPlayback = AVFoundationPlayback(options: [:])
+                        avfoundationPlayback.render()
```
BruunoMM, line 16: I've seen this line being added to a lot of tests, do you think we could create a `beforeEach` closure and move things like this to that closure?

torives: Done in f223b981

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -155,15 +155,15 @@ open class AVFoundationPlayback: Playback {
 
     open override var position: Double {
         if isDvrAvailable, let start = dvrWindowStart,
```
BruunoMM, line 157:
```suggestion
        if let start = dvrWindowStart,
```
We don't need to check that here, cause it's checked inside the `dvrWindowStart` property.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -155,15 +155,15 @@ open class AVFoundationPlayback: Playback {
...
-            let position = player?.currentItem?.currentTime().seconds {
```
BruunoMM, line 158: Maybe we could rename this to make it less confusing? It's a bit misleading cause the currentTime is not the position itself (we still need to subtract the DVR window start).

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -225,25 +243,16 @@ open class AVFoundationPlayback: Playback {
...
         trigger(.willPlay)
         if state == .stalling { trigger(.stalling) }
-        player?.play()
+        player.play()
         updateInitialStateIfNeeded()
     }
 
     private func updateInitialStateIfNeeded() {
-        guard player?.currentItem?.isPlaybackLikelyToKeepUp == true else { return }
+        guard player.currentItem?.isPlaybackLikelyToKeepUp == true else { return }
```
BruunoMM, line 255:
```suggestion
        guard let item = player.currentItem, item.isPlaybackLikelyToKeepUp else { return }
```

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
```diff
@@ -259,66 +268,29 @@ open class AVFoundationPlayback: Playback {
         return isDvrAvailable && isEpochInsideDVRWindow(liveStartTime)
     }
 
-    private func createPlayerInstance(with item: AVPlayerItem) {
-        player = shouldLoop ? AVQueuePlayer(playerItem: item): AVPlayer(playerItem: item)
-        if let player = player as? AVQueuePlayer {
-            playerLooper = AVPlayerLooper(player: player, templateItem: item)
+    private func createAVPlayer(with asset: AVAsset? = nil, shouldLoop: Bool) -> AVPlayer {
```
BruunoMM, line 271:
```suggestion
    private func createAVPlayer(with asset: AVAsset? = nil) -> AVPlayer {
```
`shouldLoop` is a property that checks a specific key in the `options` (another property), so we don't need it as a parameter.

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
torives: Done in 7c88573060002f47bf0976706b4c592bfa0b8866

Sources/Clappr/Classes/Plugin/Playback/AVFoundationPlayback.swift
torives: Actually, my initial idea was to remove this property, I don't think it makes sense to keep it as it is used only once. And in doing so (af5c2bedf999e5b6ce6140f4e56c1521de0fe935), as the option is not a Bool but an Any?, I chose to pass `shouldLoop` as a parameter to prevent the function from handling this.

BruunoMM (approved):

VitorFerraz (approved):

PR #437, opened 2021-04-01
Fix crash on SeekBubble 

daversilva:
### Goal
This fixes a bug when the `SeekBubble` is called after parent view is dismissed.

### How to test
Try seek to left or right and dismiss Player (fast way)

### DIFF
--- a/Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
+++ b/Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
@@ -36,15 +36,15 @@ class SeekBubble: UIView {
         UIView.animate(withDuration: ClapprAnimationDuration.seekBubbleShow, animations: {
             self.alpha = 1.0
             self.addRoundedBorder(with: self.bubbleHeight.constant / 2)
-            self.parentView?.layoutSubviews()
+            parentView.layoutSubviews()
         }, completion: { _ in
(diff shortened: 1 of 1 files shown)


Sources/Clappr_iOS/Classes/Animation/SeekBubble/SeekBubble.swift
```diff
@@ -27,35 +27,37 @@ class SeekBubble: UIView {
...
+        
         parentView.bringSubviewToFront(self)
         bubbleWidth.constant = parentView.frame.width
         bubbleHeight.constant = parentView.frame.height * 1.8
         
         UIView.animate(withDuration: ClapprAnimationDuration.seekBubbleShow, animations: {
             self.alpha = 1.0
             self.addRoundedBorder(with: self.bubbleHeight.constant / 2)
-            self.parentView.layoutSubviews()
+            self.parentView?.layoutSubviews()
```
tuliohmendes, line 39: Do you think it make sense to use the `parentView` reference that we just checked using `guard let` to avoid unnecessary null checks, and do the same in the other methods?

daversilva: It does make sense, thanks for suggestion. 😀

tuliohmendes (approved):

geisonfgf (approved):
Após fazer o ajuste sugerido pelo @tuliohmendes me parece tudo ok.

